Issue #289, opened 2014-09-13
Unsafe statement written to the binary log


RealRancor:
### Steps to reproduce
1. Enabling music app 0.3.4
2. Watch mysql error logfile

### Expected behaviour
No warnings about unsafe statement written to the binary log should be shown in the mysql error log

### Actual behaviour
Warnings about unsafe statement written to the binary log are shown in the mysql error log

### Server configuration
**Operating system**: Debian wheezy

**Web server:** NginX 1.6.x

**Database:** MariaDB 10.0.x

**PHP version:** 5.5.x

**ownCloud version:** 7.0.2

**Updated from an older ownCloud or fresh install:** Updated from OC6

**List of activated apps:** Default apps + mozilla_sync + music

**The content of config/config.php:**

```
<?php

define("DEBUG", true);

$CONFIG = array (
  'instanceid' => 'xxx',
  'passwordsalt' => 'xxx',
  'datadirectory' => '/var/www/owncloud/data',
  'dbtype' => 'mysql',
  'version' => '7.0.2.1',
  'dbname' => 'xxx',
  'dbhost' => 'localhost',
  'dbtableprefix' => 'oc_',
  'dbuser' => 'xxx',
  'dbpassword' => 'xxx',
  'installed' => true,
  'forcessl' => true,
  'maintenance' => false,
  'mail_smtpmode' => 'php',
  'trusted_domains' =>
  array (
    0 => 'example.com',
  ),
  'loglevel' => '0',
);
```

**Are you using external storage, if yes which one:** no

**Are you using encryption:** no

#### mysql error log
```
 [140912 11:00:05] 140912 11:00:05 [Warning] Unsafe statement written to the binary log using statement format since BINLOG_FORMAT = STATEMENT. Statements writing to a table with an auto-increment column after selecting from another table are unsafe because the order in which rows are retrieved determines what (if any) rows will be written. This order cannot be predicted and may differ on master and the slave. Statement: DELETE FROM `oc_music_tracks` WHERE `file_id` NOT IN (SELECT `fileid` FROM `oc_filecache`)

[...]

Issue #90, opened 2013-09-10
Unusable for large collections


RandolfCarter:
I know, the app is still in a preview state, but I wanted to give it a try; here's my report that alas, with a big music collection, the app is unusable at the moment.

On first access to the app, the page loading takes forever (actually it never finishes). I heard the hard drive being accessed so I thought I'd let it run. Then the access to the hard drive stopped suddenly. The server log then showed a segmentation fault in some child process; the apache and mysql process still showed high CPU usage. Only a restart of apache helped restore the ownCloud interface. And when trying to go again to the music app, the page loads forever, this time without any hard drive access, again locking up apache and mysql.

Please realize that people will have more than a few songs, and that therefore
1. Progress report during scan is crucial
2. The whole scan possibly cannot fit into one script run, or might be aborted in the middle, so there has to be a way to (ideally automatically) resume the scan
3. The scan not finishing should not break the music app; and even more importantly, should not prevent any other ownCloud access by leaving apache/mysql running at high CPU cycles...

Environment:
Ubuntu Server 12.04
ownCloud 5.0.10
music: tried with latest master (65f4678) and "various-fixes" branch (d75a4ff)
Client: Firefox 23.0 running on Ubuntu 13.04

HLFH:
@MorrisJobke @butonic Is there a way to merge https://github.com/libasys/audios and ownCloud Music? ownCloud Music has a working Ampache backend but is unusable for large collections, Audios ownCloud plugin is usable for large collections.

MorrisJobke:
We plan to join forces.

pellaeon:
@MorrisJobke Any updates? How do you plan to solve problems with large collection? It's really a pain for me, so I want to help.
Currently I find 2 bottlenecks, one is `/api/collection` and another `/api/scan?dry=0`, both seems to be caused by slow database queries. Perhaps it could be solved by memcache?

arpee:
Would also like to *bump* on this issue.

I get unresponsive pages since the application finished scanning my collection. 
(Stuck with a loading gif.)

tetzlav:
+1 
unusable slow with 10000+ songs... :(

HLFH:
@tetzlav You should use [Plex](https://plex.tv/) or [Emby](http://emby.media/) for multimedia (movies, tv shows, music) as @MorrisJobke [does](https://forums.plex.tv/discussion/108399/random-crashes-double-free-or-corruption-restart-fixes-this). For this moment, it's the better way.

[...]

PR #347, opened 2015-02-17
use IN and JOIN instead of NOT IN and simple subselect

butonic:
The background job keeps burning CPU cycles on my poor ATOM board...

Before:
```sql
# EXPLAIN DELETE FROM "oc_music_tracks" WHERE "file_id" NOT IN (SELECT "fileid" FROM "oc_filecache");
                                      QUERY PLAN                                       
---------------------------------------------------------------------------------------
 Delete on oc_music_tracks  (cost=0.00..33871909.50 rows=13900 width=6)
   ->  Seq Scan on oc_music_tracks  (cost=0.00..33871909.50 rows=13900 width=6)
         Filter: (NOT (SubPlan 1))
         SubPlan 1
           ->  Materialize  (cost=0.00..2331.24 rows=42216 width=4)
                 ->  Seq Scan on oc_filecache  (cost=0.00..1955.16 rows=42216 width=4)
```
after
```sql
# EXPLAIN DELETE FROM "oc_music_tracks" WHERE "file_id" IN (SELECT "file_id" FROM "oc_music_tracks" LEFT JOIN "oc_filecache" ON "file_id"="fileid" WHERE "fileid" IS NULL);
                                                 QUERY PLAN                                                  
-------------------------------------------------------------------------------------------------------------
 Delete on oc_music_tracks  (cost=4010.11..4018.41 rows=1 width=18)
   ->  Nested Loop  (cost=4010.11..4018.41 rows=1 width=18)
         ->  HashAggregate  (cost=4010.11..4010.12 rows=1 width=16)
               ->  Hash Anti Join  (cost=2689.86..4010.11 rows=1 width=16)
                     Hash Cond: (public.oc_music_tracks.file_id = oc_filecache.fileid)
                     ->  Seq Scan on oc_music_tracks  (cost=0.00..598.00 rows=27800 width=10)
                     ->  Hash  (cost=1955.16..1955.16 rows=42216 width=10)
                           ->  Seq Scan on oc_filecache  (cost=0.00..1955.16 rows=42216 width=10)
         ->  Index Scan using music_tracks_file_id_idx on oc_music_tracks  (cost=0.00..8.27 rows=1 width=10)
               Index Cond: (file_id = public.oc_music_tracks.file_id)
```

@MorrisJobke ping me if you want me to explain the SQL changes. I left the other three deletes as an exercise for you ... and because I need to sleep to get better.

ref #90

### DIFF
--- a/utility/helper.php
+++ b/utility/helper.php
@@ -28,8 +28,8 @@ class Helper {
 	 */
 	public function cleanUp() {
 		$sqls = array(
-			'UPDATE `*PREFIX*music_albums` SET `cover_file_id` = NULL WHERE `cover_file_id` IS NOT NULL AND `cover_file_id` NOT IN (SELECT `fileid` FROM `*PREFIX*filecache`);',
-			'DELETE FROM `*PREFIX*music_tracks` WHERE `file_id` NOT IN (SELECT `fileid` FROM `*PREFIX*filecache`);',
+			'UPDATE `*PREFIX*music_albums` SET `cover_file_id` = NULL WHERE `cover_file_id` IS NOT NULL AND `cover_file_id` IN (SELECT `cover_file_id` FROM `*PREFIX*music_albums` LEFT JOIN `*PREFIX*filecache` ON `cover_file_id`=`fileid` WHERE `fileid` IS NULL);',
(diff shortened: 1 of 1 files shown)


MorrisJobke:
:+1:

RealRancor:
@butonic 
Could it be possible that this also fix https://github.com/owncloud/music/issues/289 ?

butonic:
It avoids the subselects, so I'd say yes, but I am pretty tired and might be wrong...

RealRancor:
Thanks for the follow-up. :+1: Will have a look after the changes here was included in the next version of the app.

Issue #344, opened 2015-02-14
Music app is not allowing selection of folder to scan (personal settings)


timlepes:
**Software versions:**
music **v0.3.4**
owncloud **v8.0**
KXStudio (based on Ubuntu) **14.04.1 LTS**

**Steps to repeat:**
1. Log in to owncloud instance.
2. Select the item "Personal" under the *user-name* menu in the upper-right of the web UI.
3. Under Music find the item "Path to your music collection:".
4. Click in the path box to bring up a file-picker dialog.
5. Select a folder by clicking on it.

**Expected Result:**
Folder is opened, allowing you to **choose** it.

**Actual result:**
File-picker does not allow you to select any folders.  So you can't **choose** anything!

Ideas?  I did try to find prior incidents to no avail.  Thanks!

danmac-uk:
Hi @timlepes you may want to try the master version, v0.3.4 does not work for myself and others with ownCloud 8 but I can confirm the latest master is working great, including the functionality you are struggling with.

MorrisJobke:
Last time I tried this the file picker worked. Let me have a look at the filepicker in core in the next days. Thanks for the report.

cbxk1xg:
Same here on an nginx machine.

cbxk1xg:
I found the problem. It is in **line 28** in the file *./apps/music/js/public/settings-user.js*
```javascript
$path.on('click focus', function() {
```
The problem is the **focus** event. By clicking in the text-field you do both, focussing and clicking. The error can simply be resolved by removing the **focus** event.
```javascript
$path.on('click', function() {
```

Please test this patch and add it to the main branch of your plugin.

Best regards from good cold germany,
Thomas

cbxk1xg:
Hello?

timlepes:
Thanks, @cbxk1xg, I will try your change on my installation.  Regarding the master version from git mentioned by @danmac-uk, I will have to look into trying that.

timlepes:
@cbxk1xg  -- Thanks, Thomas!  Your change fixed it for me!

I may still try the master version to see what's different.  @MorrisJobke - I presume master is the unstable / dev branch?

Thomas' work-around works for me.  But I guess I should not close the bug until it is patched?  I'll be happy to test if needed.

Cheers!

MorrisJobke:
will be fixed in the next version.

PR #348, opened 2015-02-19
bug fix - path picker now opens only once

cbxk1xg:
Look here: https://github.com/owncloud/music/issues/344


MorrisJobke:
Thanks for the fix. Works :+1:

PR #411, opened 2015-08-14
IDBConnection does not contain "prepareQuery"

roha4000:
The IDBConnection interface does not contain a "prepareQuery" method which leads to a "Method not found" error. Using "prepare" instead.


MorrisJobke:
@roha4000 Thanks for the contribution. Which owncloud server version do you use?

roha4000:
I got this error with a 7.0.7, 8.0.6 and 8.1.1 instance all using a mySQL database

MorrisJobke:
Tested and works :+1:

PR #425, opened 2015-09-05
Protect run variables

DavidPrevot:
Follow AngularJS advices, so that any minifier can handle them properly.


MorrisJobke:
Thanks :+1:

PR #459, opened 2015-12-10
fix url and input field causing horizontal scrollbar

jancborchardt:
This fixes the horizontal scrollbar on the personal settings page when the Music app is enabled.

Please review @owncloud/designers @MorrisJobke


MorrisJobke:
Tested and works :+1: 

CSS only -> merge

PR #462, opened 2016-01-26
Improve cleanup SQL statements

roha4000:
The cleanup SQL statements took about 4-5 minutes in our production environment. After changing it to use JOINs instead the queries are done in seconds.

I also added a bit more testdata, so the integration test actually has some orphaned data to delete in cleanup method. Succesfully tested with OC 8.2.2


MorrisJobke:
Failing CI is only because of pgsql issue and not related.

MorrisJobke:
Queries make sense :+1:

MorrisJobke:
Thanks for the contribution

PR #485, opened 2016-04-02
Expose Ampache Token Generation API

gnarula:
This would help integrate the music plugin with clients that wish to use the Ampache API.

For more information, please refer to the comments at the following [doc](https://docs.google.com/document/d/189ESSu-E4kP6sXru3lxTohzXF7c8-BTDwosvtv951eI/edit?disco=AAAAAkgsNi0#heading=h.o5puy7ax05dt)

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,6 +68,53 @@ This is the common path. Some clients append the last part (`server/xml.server.p
 To use Ampache you can't use your ownCloud password. Instead, you need to generate APIKEY for Ampache.
 Go to "Your username" → "Personal", and check section Music/Ampache, where you can generate your key. Enter your ownCloud username and the generated key as password to your client.
 
+You may use the `/settings/userkey/generate` endpoint to programatically generate a random password. The endpoint expects two parameters, `length` (optional) and `description` (mandatory) and returns a JSON response.
+Please note that the minimum password length is 10 characters.
+
+```
+POST /settings/userkey/generate
+```
+
+Parameters:
+
+```
+{
+    "length": <length>,
+    "description": <description>
+}
+```
+
+Response (success):
+
+```
+{
+    "success": true,
+    "id": <userkey_id>,
+    "password": <random_password>,
+    "description": <description>
+}
+```
+
+Response (no description provided):
+
+```
+{
+    "success": false,
+    "message": "Please provide a description"
+}
+```
+
+Response (error saving password):
+
+```
+{
+    "success": false,
+    "id": null,
+    "password": <random_password>,
+    "description": <description>
+}
+```
+
 ### Installation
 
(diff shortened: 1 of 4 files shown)


controller/ampachecontroller.php
```diff
@@ -501,4 +502,26 @@ protected function get_cover() {
...
 		}
 	}
+
+	/**
+	 * this is not a part of ampache specification
+	 * This is an exposed API endpoint.
+	 * Client uses settings#addUserKey instead
+	 * @NoAdminRequired
+	 * @NoCSRFRequired
+	 * @CORS
```
MorrisJobke: I would propose to not expose another API endpoint, but simply add the annotations to the existing one in https://github.com/owncloud/music/blob/acfe12afcd05447fa0dfc698b5ff5c7f930ef552/controller/settingcontroller.php#L73

Then you also don't need to add any route ;)

gnarula: I tried to do the same initially but turns out an action annotated with `@CORS` is not allowed to be used in conjunction with session authentication https://github.com/owncloud/core/blob/master/lib/private/appframework/middleware/security/corsmiddleware.php#L82

MorrisJobke: > I tried to do the same initially but turns out an action annotated with @CORS is not allowed to be used in conjunction with session authentication

But we use the OCS API for example also in the core sharing dialog. @LukasReschke Do you have an idea how to accomplish this?

MorrisJobke: > Do you have an idea how to accomplish this?

Without code duplication.

Workaround: Create a second route in the `SettingController` and call in the method code just the original method to not copy the code but only change the authentication.

MorrisJobke:
@LukasReschke Could I get your opinion on this? :)

MorrisJobke:
I'm just thinking about this whole setUserKey method. I guess it would be better to have a "generateNewUserKey" and send it back. Otherwise this could lead to bad and weak passwords.

gnarula:
> I guess it would be better to have a "generateNewUserKey" and send it back

Should the length of the desired password be an input in that case? Maybe we can set a minimum length (say 8?) too.

Also, is it safe to assume [openssl_random_pseudo_bytes](http://php.net/openssl_random_pseudo_bytes) would be available for generating cryptographically random data?

gnarula:
Updated the PR. Implemented generateUserKey() in SettingsController

MorrisJobke:
Looks really nice 👍 

Could you add a comment about this API to the readme: https://github.com/owncloud/music#authentication That would be really nice

MorrisJobke:
Tested and works ... sorry for the long delay 👍

MorrisJobke:
I just noticed some minor stuff. I will open an additional PR for this :)

PR #493, opened 2016-04-28
Use HTTP response codes for Ampache password API

MorrisJobke:
I just switched from the success code inside the data to a HTTP return code. It's just easier to distinguish then the results ;)

Beside that: Thanks @gnarula! :)

### DIFF
--- a/controller/settingcontroller.php
+++ b/controller/settingcontroller.php
@@ -13,6 +13,7 @@
 namespace OCA\Music\Controller;
 
 use \OCP\AppFramework\Controller;
+use OCP\AppFramework\Http;
 use \OCP\AppFramework\Http\JSONResponse;
 use \OCP\Files\Folder;
 use \OCP\IConfig;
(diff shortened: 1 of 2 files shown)


controller/settingcontroller.php
```diff
@@ -13,6 +13,7 @@
 namespace OCA\Music\Controller;
 
 use \OCP\AppFramework\Controller;
+use OCP\AppFramework\Http;
```
benediktg: isn't here a leading '\' missing so that the line instead should be written like
```php
use \OCP\AppFramework\Http;
```
?

MorrisJobke: Not needed. use statements are always absolute (and this is what PHPStorm added ;) I also tested this...but maybe I need to add some more tests.

gnarula:
This is better indeed. Tried it out locally and it works well :+1:

gnarula:
Sending a POST request to the endpoint without Authentication Headers returns HTTP 500 instead of an HTTP 401. Apparently libsoup sends the authentication details only after receiving an HTTP 401 https://developer.gnome.org/libsoup/stable/libsoup-client-howto.html (refer to Handling Authentication section).

PR #501, opened 2016-06-03
Action not supported message

AshotN:
I think giving a more understandable error message will be more helpful to developers.

### DIFF
--- a/controller/ampachecontroller.php
+++ b/controller/ampachecontroller.php
@@ -110,7 +110,7 @@ class AmpacheController extends Controller {
 			case '_get_cover':
 				return $this->get_cover();
 		}
-		throw new AmpacheException('TODO', 999);
+		throw new AmpacheException('Action not supported', 501);
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


morsik:
Shouldn't be `405`? https://github.com/ampache/ampache/wiki/XML-API-ERRORS

AshotN:
I wasn't looking at the Ampache error codes, I was referencing HTTP error codes https://en.wikipedia.org/wiki/List_of_HTTP_status_codes.

morsik:
@AshotN: see this: https://github.com/owncloud/music/blob/0c0e1fadbe31a43f9c98d6a2a7b3960ff881ca0e/middleware/ampachemiddleware.php#L101

Second parameter in `AmpacheException` is not HTTP status code, but Ampache status code.

AshotN:
Changed it, but I have to say it seems odd to return what seems like an HTTP code as an Ampache code.

paulijar:
@gnarula This change seems justifiable to me. Is there any reason why this hasn't been merged for 6 months?

gnarula:
Nope, looks good to be merged

PR #503, opened 2016-06-09
Use Album Artist tag

gnarula:
Categorizes the collection based on the Album Artist tag on frontend
Updated the Ampache API to consider the tag. In case the tag is missing
a default value of "Unknown Artist" is substituted instead

I've updated the schema to hold some MusicBrainz tags but I guess it's implementation can be covered in a separate PR

### DIFF
--- a/appinfo/app.php
+++ b/appinfo/app.php
@@ -35,7 +35,8 @@ $c = $app->getContainer();
 
 // TODO: this is temporarily static because core jobs are not public
 // yet, therefore legacy code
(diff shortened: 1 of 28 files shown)


MorrisJobke:
Generally this code looks good, sadly I haven't tested it

MorrisJobke:
👍 from my side

MorrisJobke:
Let's retrigger travis. There is usually only the Postgres job that fails due to the ordering of special characters (something like á vs a)

PR #508, opened 2016-06-24
Storing extracted disc number

Faldon:
Database Updates
==============
- Updated database schema to store the disc number as integer field in albums table
- Increased patch version number to detect updating is necessary
- Added update script to truncate the database tables
- Modified album entity and mapper to reflect the new disc number field

Utility Updates
===========
- Modified scanner class to look for disk tag when parsing tracks
- Added extraction of albumartist tag
- Added alternate tag names when determining track number

Unit Test Updates
==============
- Changed unit tests to pass with the added disc number field

Feature Test Updates
================
- Changed the ampache api feature test.<br /> _Note: It seems to pass only mysql, because of different autoindices_

Travis CI
=======
- Made changes to comply with older oc installations (8.1 and before used lower case path names)

@MorrisJobke Still learning Travis -.- Hoping not to mess up too much

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,14 +29,10 @@ matrix:
       env: DB=mysql OWNCLOUD=stable7
   fast_finish: true
 
-branches:
-  only:
-    master
-
 before_install:
   - cd ..
   - git clone https://github.com/owncloud/core.git --depth 1 -b $OWNCLOUD
-  - mv music core/apps/
+  - mv music core/apps/music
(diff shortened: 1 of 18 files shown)


gnarula:
Thanks for your PR! :+1:  I've made a few changes, considering there was a merge conflict after #503 was merged into master. The schema's changed a bit and album artists are read from the TPE2 tag ('band' in getID3).

PR #509, opened 2016-06-24
Extracting album art and use of album artists

Faldon:
Controller Updates
==============
- Try to extract a cover art from the first track of an album if no cover file was specified
- Set album artist as artist for displaying tracks where artist != albumartist
- Modified matching entity to always return a link to the cover file

Frontent Updates
=============
- Modified AngluarJS directive albumart to load cover asynchronously and on failure generate the generic albumart

Other
====
- Adopted changes for Travis and Unit testing from #508


.travis.yml
```diff
@@ -62,6 +62,13 @@ before_install:
...
   - bash -c "if [ '$DB' == 'pgsql' ]; then psql -U postgres -d oc_autotest -w -c \"INSERT INTO oc_music_ampache_users (user_id, hash) VALUES ('ampache', '3e60b24e84cfa047e41b6867efc3239149c54696844fd3a77731d6d8bb105f18');\"; fi"
   - cd apps/music
+  # Lowering camel case lib paths in autoload for older oc installations
+  - bash -c "if [ '$OWNCLOUD' != 'master' ]; then sed -i '24 s/ str_replace(/ strtolower(str_replace(/' tests/php/autoloader.php; fi"
+  - bash -c "if [ '$OWNCLOUD' != 'master' ]; then sed -i '31 s/ str_replace(/ strtolower(str_replace(/' tests/php/autoloader.php; fi"
+  - bash -c "if [ '$OWNCLOUD' != 'master' ]; then sed -i '38 s/ str_replace(/ strtolower(str_replace(/' tests/php/autoloader.php; fi"
+  - bash -c "if [ '$OWNCLOUD' != 'master' ]; then sed -i '24 s/;/\);/' tests/php/autoloader.php; fi"
+  - bash -c "if [ '$OWNCLOUD' != 'master' ]; then sed -i '31 s/;/\);/' tests/php/autoloader.php; fi"
+  - bash -c "if [ '$OWNCLOUD' != 'master' ]; then sed -i '38 s/;/\);/' tests/php/autoloader.php; fi"
```
gnarula: This would break if we add another build for versions >= 9.1 (where PSR-4 changes were made). Not required now anyway since `tests/php/autoloader.php` now fall backs to lowercase path in case of failure

controller/apicontroller.php
```diff
@@ -107,6 +107,11 @@ public function collection() {
 		$artists = array();
 		foreach ($allTracks as $track) {
 			$artist = &$allArtistsById[$track->getArtistId()];
+			$albumArtists = $this->albumBusinessLayer->find($track->getAlbumId(), $this->userId);
+
+			if($albumArtists->getArtistIds()[0]!=$artist['id']) {
+				$artist = &$allArtistsById[$albumArtists->getArtistIds()[0]];
+			}
```
gnarula: Again, Album Artist is a separate tag altogether and its value should not be taken from the Artist Tag. This would result in a wrong value for an Album containing collection of songs of various artists. Please refer to #503 for the changes

gnarula:
Thanks for the PR :) Could you please address the things I've mentioned in the comments and update the PR after resolving the merge conflicts?

I don't think the commit for the unit test and fixing travis builds is required anymore since #503 fixed them already

Faldon:
I will then fetch back the .travis.xml from master branch and do the required changes.

Faldon:
So. I fetched back the .travis.yml and the apicontroller.php and rewrote the cover function, again trying to extract the album art from first track of album if no cover file for the album was given.

gnarula:
Looks good to me! Could you resolve the merge conflicts and squash commits into one please? Changes in phpunit.xml and autoloader.php aren't required

Thanks once again for working on this! :)

Faldon:
Your welcome.
I fetched back your master tree and checked out the files not needing to change. I than squashed the commits.
I will initiate a push now. I wasn't in need for such a 'complex' operation on git before, so let me know if it is more comfortable for you, if I open a new PR.

gnarula:
:( still showing some old changes. Try creating a new PR if that makes things easier. Just make sure you branch off from the HEAD of music:master and then add your changes and it should be good to go

PR #510, opened 2016-06-24
Ampache XML API

Faldon:
- Added missing album action for ampache api
- Implemented missing limit and offset parameters
- Implemented missing add and update parameters
- Changed albums xml template to get disc number from database, as introduced with #508 

- Adopted changes for travis and unit tests from #508 rsp. #509

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,6 +29,10 @@ matrix:
       env: DB=mysql OWNCLOUD=stable7
   fast_finish: true
 
+branches:
+    only:
+        - master
+
 before_install:
   - cd ..
   - git clone https://github.com/owncloud/core.git --depth 1 -b $OWNCLOUD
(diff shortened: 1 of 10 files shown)


gnarula:
Is it a good idea to use `mtime` as the "addition" time? Even when considering it as update time, the AmpacheAPI doesn't provide updated information (in terms of song metadata) yet. This would probably beat the purpose of using these parameters for filtering queries by clients.

Maybe we should keep a track of the first time a track was scanned. On modification, perhaps we should initiate a scan again and store the last time at which file was scanned in a separate column. Thoughts?

Faldon:
When I began working on the Ampache API, I just wanted it a bit more compatible to Ampache clients. :-) When confronting the hook you mention, I just took the mtime to produce a more or less reliable information.
But I second that, that we should add more information to the database, at least considering ampache api.

PR #512, opened 2016-07-13
Added extraction of album art

Faldon:
This is the new PR replacing #509 
It now should only contain the changes necessary to extract the album art from the first track of the album if no cover was specified.

Should close #509


gnarula:
Looks good! I noticed that fetching the album art fails at times (getID3 complains of invalid APIC tag when accessing the resource via "oc://" protocol) but that's unrelated to this PR since it persists when we scan for music as well. We can perhaps use the thumbnails generated by core (`Preview/MP3`) to avoid this altogether. Unfortunately, that API isn't exposed publically yet :(

For now, this is a welcome addition indeed :+1: Thank you for your PR :smile: Merged

Issue #434, opened 2015-10-20
Blank page after OC 8.2


briskspirit:
Upgraded to OC 8.2 from 8.1.3 and your app stopped to work, just see blank page. Tried to reinstall and this doesn't helped..

MorrisJobke:
Is there anything in the owncloud.log file in the data dir?

briskspirit:
nope, even with loglevel set to 0

briskspirit:
So any ideas how to catch this error if logs are empty?..

MorrisJobke:
> So any ideas how to catch this error if logs are empty?..

Anything the the browsers JS console? Or in the network tab of the browsers web development console. Maybe there is a request that fails (should also be possible to spot in the access log of the webserver).

Or in the error log of the web server.

briskspirit:
Will try to catch today and write back.

MorrisJobke:
@briskspirit Awesome :) Thanks

briskspirit:
apache2 log is ok
console gave js error:
Error: [$injector:unpr] http://errors.angularjs.org/1.2.21/$injector/unpr?p0=eProvider%20%3C-%20e

And this(on all pages)
Failed to load resource: the server responded with a status of 404 (Not Found)
https://xxx/assets/jquery.min.map

MorrisJobke:
> And this(on all pages)
> Failed to load resource: the server responded with a status of 404 (Not Found)
> https://xxx/assets/jquery.min.map

That one can be ignored.





> console gave js error:
> Error: [$injector:unpr] http://errors.angularjs.org/1.2.21/$injector/unpr?p0=eProvider%20%3C-%20e

This looks interesting. Just to verify: You don't use a JS blocker, right?

briskspirit:
Nope, no blockers. And all worked fine before upgrade to OC 8.2

MorrisJobke:
> Nope, no blockers. And all worked fine before upgrade to OC 8.2

That's even weirder. Because this is the JS part only. What apps do you have installed?

briskspirit:
Anything else I can do/check?

MorrisJobke:
> Anything else I can do/check?

I don't know anything yet :(

pedronariyoshi:
Same thing here. Even same JS error. Besides the official apps, I have: Archive, Documents, Music and ODF reader enabled.

briskspirit:
Oh, about my list of apps - tried to shut off ALL of them, leaving just Music app ))

[...]

PR #513, opened 2016-07-14
Fixed the "Blank page with asset-pipeline.enabled" bug #434

paulijar:
- Minification broke the injection because annotation was missing

### DIFF
--- a/js/app/directives/albumart.js
+++ b/js/app/directives/albumart.js
@@ -19,7 +19,7 @@
  *
  */
 
-angular.module('Music').directive('albumart', function($http) {
+angular.module('Music').directive('albumart', ['$http', function($http) {
 	return function(scope, element, attrs, ctrl) {
 		var setAlbumart = function() {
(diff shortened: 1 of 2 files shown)


Gomez:
Works! :+1: 
@gnarula Is this good to merge? 

@paulijar Congrats! Great work!

gnarula:
Album arts break for me when I enable asset-pipelining :( @Gomez @paulijar do you guys get a traceback like this in the console https://paste.gnome.org/pgsjmbcln ?

gnarula:
I think I got it. There's an annotation missing in the albumart directive (`$http`). @paulijar could you please update `js/public/app.js` and `js/app/directives/albumart.js` and update the PR?

Thanks for your PR once again :)

paulijar:
I figured out the same just before reading your latest comment :). For some reason, the albumart problem was not present when I was originally testing this on top of the release version (0.3.11) of the music app. But with the latest sources from github the problem was there.

PR #514, opened 2016-07-15
AmpacheAPI: fix add and update timestamp in handshake endpoint

gnarula:
The music_tracks table now stores the time at which a track was
added and updated in the music app

update tag in the handshake endpoint contains the max of added
and updated column whichever is the latest. add tag in the handshake
endpoint contains the max of added column


paulijar:
This feature has now been implemented in Music v1.0.0. The implementation is technically a bit different than the one proposed here, but gets the same thing done.

PR #519, opened 2016-07-26
Fixed playback order when playing album

paulijar:
The playback order of tracks within an album was odd, to say the least. Tracks were being skipped over and sometimes the same track could be played many times, and all this while neither Shuffle nor Repeat was active. The problem was not visible when manually changing to next or previous track with the buttons, but only when the tracks were actually played through and the next track was started automatically.

The bug was caused by registering the same callbacks again and again for the same PlayerWrapper instance every time when the playing song changed. Hence, the $scope.next() function got called many times when the track ended.

### DIFF
--- a/js/app/controllers/playercontroller.js
+++ b/js/app/controllers/playercontroller.js
@@ -39,6 +39,34 @@ angular.module('Music').controller('PlayerController',
 		total: 0
 	};
 
+	$scope.player.on('buffer', function (percent) {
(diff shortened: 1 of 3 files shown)


paulijar:
So @gnarula, @MorrisJobke, would you consider merging this one? As I see it, this is a kind of critical bug.

gnarula:
Looks good, sorry for taking so long to review these :+1: . I was moving back to college.

PR #517, opened 2016-07-21
Fixed performance issues related to cover image loading

paulijar:
The UI performance was absolutely horrible when the library contained lots of albums with embedded cover art. The biggest problem was that all the cover art was reloaded each time when a track or album or artist was selected for playing. The playing doesn't start before the covers are loaded. Another problem was that each cover picture was actually loaded twice.

PR #520, opened 2016-07-26
Allow playback to start before all the album cover arts are loaded

paulijar:
After the performance fixes made in PR #517, there was still one problem in the perceived performance: When loading the Music app, all the cover images had to be completely loaded before playing the music was possible. This was kind of bad since playing music is the primary function of the application and seeing the cover images is more or less a nice-to-have feature. 

This PR addresses the issue by running the HTTP GET requests to retrieve the cover images via a queue. This way, browser can serve other requests (like downloading the music file) between getting the individual cover images.

The queued requests are not run in purely sequential order because this would increase the total time taken to download the cover images. Instead, at maximum 5 simultaneous requests are allowed. With this setup, there is no measurable difference in the total time taken to download the images but music playback can still be started with rather short delay. To easily implement this kind of "semi-sequential" behavior, the library ngQueue was added to the project.


js/app/directives/albumart.js
```diff
@@ -19,7 +19,13 @@
...
  */
 
-angular.module('Music').directive('albumart', ['$http', function($http) {
+angular.module('Music').directive('albumart', ['$http', '$queueFactory', function($http, $queueFactory) {
+	// Calling $http.get immediately for all album cover images would be bad idea because it would
+	// block the playback until all the covers are loaded. Hence, we use queue which allows 5 simultaneous
+	// HTTP GET queries. This is faster than running them only one at a time, but still enables starting
+	// the playback with rather short delay.
+	var httpQueue = $queueFactory(5);
+
```
gnarula: trailing whitespace

.gitignore
```diff
@@ -55,6 +55,6 @@ coverage-html-integration
 # L10n
 l10n/templates/music.pot
 
-vendor/
 bin/
 tests/behat.yml
+/vendor/
```
gnarula: please use unix style line endings (LF) https://doc.owncloud.org/server/9.1/developer_manual/general/codingguidelines.html#coding

js/public/app.js
```diff
@@ -498,7 +499,13 @@ angular.module('Music').controller('PlaylistController',
...
 }]);
 
-angular.module('Music').directive('albumart', ['$http', function($http) {
+angular.module('Music').directive('albumart', ['$http', '$queueFactory', function($http, $queueFactory) {
+	// Calling $http.get immediately for all album cover images would be bad idea because it would
+	// block the playback until all the covers are loaded. Hence, we use queue which allows 5 simultaneous
+	// HTTP GET queries. This is faster than running them only one at a time, but still enables starting
+	// the playback with rather short delay.
+	var httpQueue = $queueFactory(5);
+
```
gnarula: trailing whitespace

js/public/app.js
```diff
@@ -498,7 +499,13 @@ angular.module('Music').controller('PlaylistController',
 
 }]);
 
-angular.module('Music').directive('albumart', ['$http', function($http) {
+angular.module('Music').directive('albumart', ['$http', '$queueFactory', function($http, $queueFactory) {
```
gnarula: please use LF line endings

gnarula:
Looks good, thanks for your PR! Please make the changes I've mentioned in the comments and it's good to commit :)

paulijar:
Thanks for your review, @gnarula. The findings are now fixed.

gnarula:
Thanks! Merged :)

PR #521, opened 2016-07-26
Fix on the disc number extraction

paulijar:
The code in scanner.php tried to extract the disc number from track using the key 'discnumber'. However, at least with my MP3 files the disc number is actually found with key 'part_of_a_set'. Hence, all tracks were added to the database with disc number 1.

I modified the scanner.php to accept both 'discnumber' and 'part_of_a_set'. But actually, looking at the sources of getID3, I believe that the key 'discnumber' will never work.

### DIFF
--- a/utility/scanner.php
+++ b/utility/scanner.php
@@ -239,6 +239,8 @@ class Scanner extends PublicEmitter {
 			$discNumber = "1";
 			if($hasComments && array_key_exists('discnumber', $fileInfo['comments'])){
 				$discNumber = $fileInfo['comments']['discnumber'][0];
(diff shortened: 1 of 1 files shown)


Faldon:
The discnumber tag works well with ogg comments. :-) But I have to confess, I didn't dive deep enough into the id3 specifications when adding the discnumber. Just assumed it right, for Asunder writes this tag as well as Puddletag and my players recognizes them (an exception is Music on Windows Phone. Should try part_of_a_set).
But your addition looks good to me.

gnarula:
Looks good! `module.tag.id3v2.php::FrameNameShortLookup()` maps the frame and tag name for future reference.

PR #522, opened 2016-07-27
Fixes on small UI glitches

paulijar:
This PR provides fixes for following small UI glitches:

- On Firefox, the small play icon was not visible in front of the track name while hovering/playing if the track name was truncated. Apparently, Firefox didn't tolerate the negative margin in this situation. The bug was not visible on Chrome or Edge.
- There was a tiny bit of jitter when tracks of album were hovered over. The track name moved horizontally maybe one pixel when the play icon was shown/hidden. This was visible on all browsers.
- The mouse cursor didn't correctly change to pointer icon on all clickable items like album names and the "Show all X songs..." item. This happened on all browsers.

The proper functionality of the UI was tested on up-to-date versions of Firefox, Chrome, and Edge.

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -32,11 +32,15 @@
 }
 
 .play {
-	display: none;
+	display: inline;
+	visibility: hidden;
 	/* opacity */
 	-ms-filter: "progid:DXImageTransform.Microsoft.Alpha(Opacity=70)";
 	filter: alpha(opacity = 70);
 	opacity: .7;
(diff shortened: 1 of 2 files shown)


gnarula:
Looks good! Thanks for your PR :)

jancborchardt:
@gnarula @paulijar you should cross-review the other pull requests so we can get the Music app fixed up. :)

PR #523, opened 2016-07-27
Show track artist on track when it differs from the album artist

paulijar:
The Music app had such a shortcoming that the track artist could not be seen from anywhere. Only the album artist was visible.

This PR addresses the issue by adding the track artist name after the track name in case the track artist and album artist are not the same. In addition, the track artist name is shown in the top pane for the currently playing song instead of the album artist. The albums are grouped by album artists and the album artist names are used as group titles like before.

show_track_artist

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,8 +79,8 @@ Parameters:
 
 ```
 {
-    "length": <length>,
-    "description": <description>
+	"length": <length>,
+	"description": <description>
 }
 ```
 
@@ -90,9 +90,9 @@ Response (success):
 HTTP/1.1 201 Created
 
 {
-    "id": <userkey_id>,
-    "password": <random_password>,
-    "description": <description>
+	"id": <userkey_id>,
+	"password": <random_password>,
+	"description": <description>
 }
 ```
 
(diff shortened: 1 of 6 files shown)


jancborchardt:
@MorrisJobke @gnarula @Faldon can you check this out? :)

Ideally also review each other’s pull requests: https://github.com/owncloud/music/pulls then everyone wins! :) :tada:

Faldon:
This looks good to me and I personally like the changes.

jancborchardt:
@paulijar the only thing I would change is the `//` as delimiter as it looks very technical and we don’t use it anywhere else. Instead, can you change it to simply parentheses around the artist name? :)

paulijar:
@jancborchardt Sure. I imitated the //-syntax from the foobar2000 music player, but I do remember it seemed a bit odd the first time I saw it on foobar. Now I have got used to it :). I'll give a try for parentheses when I have time.

paulijar:
Now it looks like this. Maybe it's better, at least not any worse.

artist_name_with_parentheses

gnarula:
My appologies for taking so long to review this. I had been occupied with some college work in the past few weeks.

PR looks good, thanks once again :+1:

PR #524, opened 2016-07-28
Use track artist and album artist as fallbacks for each other

paulijar:
In the current main branch implementation, all albums without the 'album artist' tag set are listed under 'Unknown Artist'. My own music library, for example, contains dozens of such albums. I have used the 'album artist' tag mostly in cases where the album artist and track artist actually differ. For the rest of the tracks, I haven't bothered since it makes no difference in majority of the music players I have used. I believe I'm not the only one with the same situation.

In this PR, I have mitigated the problem by adding the following fallback logic:
- If the music file has no 'album artist' tag, the album artist is assumed to be the same as track artist
- If the music file has no 'artist' tag, the track artist is assumed to be the same as the album artist
- If neither 'artist' nor 'album artist' tag is available, then both artist and album artist are set to 'Unknown Artist'

The logic is applied at the library scanning time. Hence, the music database has to be regenerated to benefit from the change.

The logic is heuristic, of course, and may not create perfect result in all cases. However, I believe that this is more in line with the user expectations in majority of cases than substituting 'Unknown Artist' for all missing artist fields.

### DIFF
--- a/utility/scanner.php
+++ b/utility/scanner.php
@@ -163,18 +163,24 @@ class Scanner extends PublicEmitter {
 			if($hasComments && array_key_exists('artist', $fileInfo['comments'])){
 				$artist = $fileInfo['comments']['artist'][0];
 			}
-			if($artist === '' || $artist === null){
-				// set artist to Unknown Artist
-				$artist = 'Unknown Artist';
-			}
 
 			// albumArtist
 			$albumArtist = null;
 			if($hasComments && array_key_exists('band', $fileInfo['comments'])){
 				$albumArtist = $fileInfo['comments']['band'][0];
 			}
-			if($albumArtist === '' || $albumArtist === null){
-				// set albumArtist to Unknown Artist
+
+			// use artist and albumArtist as fallbacks for each other
+			if($this->isNullOrEmpty($albumArtist)){
+				$albumArtist = $artist;
+			}
+			if($this->isNullOrEmpty($artist)){
+				$artist = $albumArtist;
+			}
+
+			// set 'Unknown Artist' in case neither artist nor albumArtist was found
+			if($this->isNullOrEmpty($artist)){
+				$artist = 'Unknown Artist';
 				$albumArtist = 'Unknown Artist';
 			}
 
@@ -184,7 +190,7 @@ class Scanner extends PublicEmitter {
 			if($hasComments && array_key_exists('title', $fileInfo['comments'])){
 				$title = $fileInfo['comments']['title'][0];
 			}
-			if($title === null || $title === ''){
+			if($this->isNullOrEmpty($title)){
 				// fallback to file name
 				$title = $file->getName();
 				if(preg_match('/^(\d+)\W*[.-]\W*(.*)/', $title, $matches) === 1) {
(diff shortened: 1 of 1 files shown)


gnarula:
> I have used the 'album artist' tag mostly in cases where the album artist and track artist actually differ

Please have a look at https://mkoby.com/2007/02/18/artist-versus-album-artist/

> If the music file has no 'album artist' tag, the album artist is assumed to be the same as track artist

Consider the case where an album is a compilation of tracks by different artists. The album artist is usually Various Artists in this case and rightfully so.

> If the music file has no 'artist' tag, the track artist is assumed to be the same as the album artist

Again, the above example holds.

Point to consider is album artist != track artist at all times and I don't think it's right to substitute one for another as a "fallback"

paulijar:
Yes, I'm aware of the differences of album artist and track artist and the intended purpose of each tag. I do have many compilation albums where I have tagged album artist as "Various artists" and track artist as the individual performed of the track in question. Those albums still work exactly the same as before after the proposed change.

The point is that sometimes the library is less than perfectly tagged. In those cases the software has to apply some fallback logic to categorize the tracks. The current fallback logic is to substitute the missing information with "Unknown Artist" which is *never* the right guess. The proposed change is to make an educated guess instead, which is quite often the right one. Most if not all music players I have ever used have this kind of logic (provided that the player utilizes 'album artist' tag at all). Try Microsoft Groove, for example.

From the point-of-view of the end-user, the question is about how much effort is needed for satisfactory results. The current implementation requires having both artist and album artist set on each and every track. With the proposed change, both tags are needed for the compilation albums but not for other albums. And if the user already has a perfectly tagged library, then nothing will change compared to the current functionality.

paulijar:
As an added point, the music app already does contain some heuristic logic related to track organization. For example, directory name is used as album name if album name is not specified. And track number is tried to be extracted from the file name if track number tag is not specified. Both of these may lead to wrong results but they are good guesses.

Faldon:
I side with @paulijar in this point. Personally I think, chances are high, that if one of the two artist information is missing, they carry the same information.
If album artist != artist, than both tags are extracted as separate information, what is a correct behaviour.
Dealing with user input (tags in this case) is always hard due to the huge range of variations one has to consider. ;-)

gnarula:
On second thought this indeed seems like a good fallback strategy and is in line with the current heuristics in the app. Thanks for your suggestions @paulijar @Faldon

paulijar:
Cool, thanks :).

lanbase:
Ok, I was happy when I read this but since the update to 3.13. I can not group the tracks of a compilation cd anymore . the album is duplicated as much as artist are on there.

I have tagged  1 name in the  "Album Artist" field but that doesn´t help, maybe it is because the albums I am talking about are FLAC and there is a problem reading the tags properly ?

no rant, just giving information.

Regards, Lanbase

gnarula:
@lanbase you're right, this is FLAC specific. We assumed that getID3 puts the Album Artist information in the 'band' key of the comments array but turns out it does so only for ID3 tags and not the APE tags which are usually found in FLAC files.

The solution is to look for 'albumartist', 'album artist' and 'album_artist' as well while determining the Album Artist (https://www.blisshq.com/music-library-management-blog/2013/08/13/mind-your-flac-albumartists/). Pushing the fix in a while :) Thanks for your input

PR #528, opened 2016-08-19
FileResponse: support range requests

gnarula:
This PR aims to introduce support for HTTP Range header in FileResponse class and would allow clients using the AmpacheAPI to seek tracks. The implementation doesn't support Multiple Ranges like `bytes=0-49,50-99` since I doubt render() supports multi-part message body.

I've tried the implementation on GNOME Music which uses gstreamer behind the scenes and seeking works fine.

### DIFF
--- a/http/fileresponse.php
+++ b/http/fileresponse.php
@@ -21,13 +21,15 @@ use OCP\AppFramework\Http;
 class FileResponse extends Response {
 
 	protected $file;
+	protected $start;
+	protected $end;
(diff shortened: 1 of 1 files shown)


MorrisJobke:
Looks good to me. I guess we can't reuse the servers implementation of the range request because we have a different authentication (amapache auth) on top.

:+1:

gnarula:
Thanks for reviewing this! Merged :)

Issue #365, opened 2015-04-13
Clicking on the song/artist/cover in the play bar should scroll the view to that album


jancborchardt:
Currently it’s a bit cumbersome to get back to the currently playing artist. Simple fix: When clicking on song/artist/cover in the play bar, scroll the main view to that album.

Yes, clicking anywhere there should just scroll to the album. We don’t need to differentiate between clicking on album or artist for now. cc @MorrisJobke

paulijar:
I have a working implementation for this feature but it would be easier to make a pull request if #523 is merged first.

paulijar:
@gnarula This feature is implemented and could be closed

PR #529, opened 2016-08-20
Fixed automatic starting of music playback via a link + scroll to the started album

paulijar:
Looking at the source codes, it was clear that music playback was supposed to start automatically when the Music app was entered by clicking a music file in the Files app. Similarly, playback was supposed to start if an URL like <server>/owncloud/index.php/apps/music/#/album/123 was entered. Neither of these features worked with the latest SW versions and I believe they have not worked for quite some time. These features are fixed with this PR.

In addition, the automatic playback is enhanced by also scrolling the view automatically to the album which began to play via a link.

As a third change, the currently playing album can now be scrolled to view by clicking the album art or song info in the controls pane. This is the feature requested in #365.

There is such a small annoyance, that when an album is scrolled to view, the title of the album and the first track is actually left under the controls pane. Similar issue is visible when scrolling with the artist name initials next to the scroll bar. It should be possible to fix this by scrolling with an offset but I couldn't make it work with short prototyping. I may return to it later. On the other hand, it might make sense to slightly change the design of the controls pane to get rid of these scrolling issues as well as the top of the scroll bar being hidden by the pane.

### DIFF
--- a/js/app/controllers/maincontroller.js
+++ b/js/app/controllers/maincontroller.js
@@ -9,8 +9,8 @@
  */
 
 angular.module('Music').controller('MainController',
-	['$rootScope', '$scope', '$route', 'ArtistFactory', 'playlistService', 'gettextCatalog', 'Restangular',
-	function ($rootScope, $scope, $route, ArtistFactory, playlistService, gettextCatalog, Restangular) {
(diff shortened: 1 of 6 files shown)


gnarula:
Looks good! I was thinking about working on this over the weekend but you beat me to it :P Thanks once again for your contribution @paulijar :)

PR #530, opened 2016-08-21
Fix regression on mobile style

paulijar:
The PR #522 had caused regression on the layout on mobile devices: the tracks were aligned incorrectly. This PR fixes the regression.


gnarula:
Ah, overlooked the mobile view! Thanks for this, merged!

PR #531, opened 2016-08-23
Fixed resizing of albumart placeholders when style changes (mobile/desktop)

paulijar:
- The album art placeholders were not resized correctly when the style sheet used got changed on window size change. The font size and the vertical alignment of the letter changed but the colored background retained its previous size.
- This bug seems to be originated from changed behavior of the imageplaceholder() function in the core by commit 6248bad0f7. This commit is first included in Owncloud 9.0.0.


gnarula:
Looks good, merged!

Issue #533, opened 2016-09-01
Is there any reason to use aurora.js to play mp3 files?


paulijar:
Music player currently uses soundmananger2 library to play .ogg files and aurora library to play .mp3 and .flac files. The former does not support .flac files and the latter does not support .ogg files but .mp3 files could be played by both. Is there any reason why aurora has been selected for .mp3?

The reason I'm asking is because aurora has a lots of problems playing mp3 files. At least these:
- Some mp3 files do not play at all (as noticed in README.md)
- There is some distortion in the playback quite often (#490). Especially if the web browser is doing anything else while playing music in one tab.
- The sound quality is totally unacceptable on Android Chrome. (https://github.com/audiocogs/aurora.js/issues/173)
- The mp3 playback does not work on Internet Explorer versions older than 11.

According to my tests, all of these would be fixed by changing the mp3 playback engine to soundmanager2. Playback would still not work on IE8, but IE9 would work fine.

I also tested that updating aurora.js and its MP3 plugin to most recent released versions does not help on any of the above mentioned problems.

paulijar:
@MorrisJobke, @pellaeon, @gnarula, anyone?

pellaeon:
Chromium has no built in codec for MP3 (but Chrome does), see https://www.chromium.org/audio-video

paulijar:
Okay @pellaeon, thanks for the info. Maybe PlayerWrapper could then detect the situation somehow and fall back to aurora.js only on Chromium?

pellaeon:
Yes, that is possible, pull requests welcomed :-D

2016-09-04 17:37 GMT+08:00 Pauli Järvinen <notifications@github.com>:

PR #535, opened 2016-09-04
Use SoundManager2 for MP3 files when possible

paulijar:
As discussed in #533, Aurora.js has some pretty bad problems playing MP3 files. We can get rid of these issues by playing MP3 files with SoundManager2 instead. The only problem is that SoundManager2 cannot play MP3 files on all browsers (Chromium) because of a missing built-in decoder.

With this PR, the logic for selecting the playback engine is altered so that PlayerWrapper asks the SoundManager2 library if it can handle the file in question. The result depends on the decoders available. If SM2 cannot be used, then PlayerWrapper falls back to Aurora.js.

The change has been tested on Firefox, Chrome, Chromium, Android Chrome, Edge, and IE9. All of these browsers can now play MP3 files. Chromium, Edge and IE9 cannot play OGG files but that was the case also before this change. IE9 cannot play FLAC files, either.

cc @pellaeon, @gnarula

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,9 +19,9 @@ _Note: sometimes it might be unable to play all FLAC and MP3 files (especially i
 
 This app utilizes 2 backend players: Aurora.js and SoundManager2.
 
-Aurora.js uses Javascript and HTML5 Audio API to decode and play music, it doesn't use the browser's built-in codec. This app choose Aurora.js for FLAC and MP3 files.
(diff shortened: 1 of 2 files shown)


pellaeon:
Looks good to me 👍 I'll wait for @gnarula to merge the PR.

gnarula:
Yup looks good to me too :+1: The improvement is remarkable in Chrome for Android! Awesome work as always @paulijar :)

PR #536, opened 2016-09-06
Fixed and enabled seeking with SoundManager2

paulijar:
Seeking was being disabled in the Music app because of an "angular issue". Angular, however, is working as it is supposed to and the bugs were in the Music app codes. After fixing these bugs, the seeking is working fine with SoundManager2.

Aurora (or its plugins), on the other hand, seems to have issues with seeking. Seeking is working properly only on some of the MP3 files. It did not work on either of the two FLAC files I tested. Hence, I thought it is better to leave the seeking disabled with Aurora.

Also some other fixes related to the progress bar were made. See the individual commit messages for full details.

### DIFF
--- a/js/app/playerwrapper.js
+++ b/js/app/playerwrapper.js
@@ -48,15 +48,26 @@ PlayerWrapper.prototype.togglePlayback = function() {
 	}
 };
 
+PlayerWrapper.prototype.seekingSupported = function() {
+	// Seeking is not implemented in aurora/flac.js and does not work on all
+	// files with aurora/mp3.js. Hence, we disable seeking with aurora.
+	return this.underlyingPlayer == 'sm2';
+};
+
 PlayerWrapper.prototype.seek = function(percentage) {
-	console.log('seek to '+percentage);
-	switch(this.underlyingPlayer) {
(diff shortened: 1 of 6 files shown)


js/app/playerwrapper.js
```diff
@@ -48,15 +48,26 @@ PlayerWrapper.prototype.togglePlayback = function() {
...
-	switch(this.underlyingPlayer) {
-		case 'sm2':
-			this.sm2.setPosition('ownCloudSound', percentage * this.duration);
-			break;
-		case 'aurora':
-			this.aurora.seek(percentage * this.duration);
-			break;
+	if (this.seekingSupported()) {
+		console.log('seek to '+percentage);
+		switch(this.underlyingPlayer) {
```
gnarula: the switch case looks redundant if seeking's only supported by SM2

paulijar:
@gnarula Well, the switch-case is redundant per-se, but on the other hand, it enables easy testing of seeking on aurora by modifying only the function seekingSupported. It is also possible that at some point we want to enable seeking for aurora with either mp3 or flac files or both.

PR #280, opened 2014-08-27
volume control

MorrisJobke:
cc @pellaeon

PR #537, opened 2016-09-10
Small and subtle volume control

paulijar:
This PR started as rebase of MorrisJobke's volume control from #280 but ended up quite a bit further.

Features:
- Works both with Aurora.js and SoundManager2 and with all file formats
- Small and subtle layout, fitting also narrow windows
- The volume setting is preserved between sessions with a cookie
- Works with Firefox, Chrome, Edge, IE11, IE10. The look-and-feel varies somewhat between the browsers as each of them implements the HTML5 range input a bit differently.
- Hidden on IE9 and IE8 (these don't support the HTML5 range input at all)
- Hidden in the mobile layout. In the tablet layout, volume control is shown but its icon is hidden to save space.

volume_control


pellaeon:
Looks good to me 👍

gnarula:
Looks good to me too! :+1: Merged :)

Issue #420, opened 2015-09-03
Update actual files 


MorrisJobke:
- #416 changed app.js which is generated -> move the changes also to the actual files.

paulijar:
This was fixed in #539

PR #539, opened 2016-09-23
Fixed the javascript build system

paulijar:
The file js/public/app.js has been originally generated from the other, module-specific .js files. Lately, however, the automatic generation of app.js has been broken, and the file has been maintained manually. This process is dull and error-prone which is proven by the fact that there have been numerous cases where contents of app.js have diverged from the "actual souces".

This PR fixes the problems of the build system and replaces app.js with the generated file. This also closes the issue #420.


pellaeon:
Thanks for your help! LGTM 👍

gnarula:
Looks good! Merged!

PR #540, opened 2016-09-26
Adjust the controls bar width to not overlap with the scroll bar

paulijar:
Fixed the scroll bar being hidden behind the controls bar. The implementation is based on the fix of similar problem in the Mail app: https://github.com/owncloud/mail/pull/874.


gnarula:
Looks good to me :+1: Merged :)

PR #542, opened 2016-09-29
Refactored scanner.php

paulijar:
Made scanner.php less redundant on how it accesses $fileInfo from getID3. This should help when replacing getID3 with ID3Parser, which generates a bit different file info structure..


gnarula:
Looks good to me, merged

PR #543, opened 2016-10-02
Fixed album deletion, Scrutinizer issues, and tests

paulijar:
- When the last track of an album was removed, the album was supposed to be removed from the database. This didn't work because the function AlbumMapper.deleteById tried to modify also table *PREFIX*music_album_artists which no longer exists in the Music app. This caused an exception, error was logged, and the empty album was never removed from the database.
- Similar problem was present when the path to music library was changed from the user personal settings. Here it didn't cause any functional failure, but an error was logged.

### DIFF
--- a/db/albummapper.php
+++ b/db/albummapper.php
@@ -13,10 +13,9 @@
 namespace OCA\Music\Db;
 
 use OCP\AppFramework\Db\DoesNotExistException;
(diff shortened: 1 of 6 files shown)


gnarula:
Nice catch! :) Merged!

PR #544, opened 2016-10-13
Improved the layout of the "New music available" and "Scanning" boxes

paulijar:
When there are already some albums in the library, and new music files become available, the "New music available" box is overlayed on top of the album view. This box used to have quite odd layout as it spanned the whole screen width while its content could fit much smaller box. Also, it overlayed the album art and the track names but not anything written in dimmed font (track numbers, album years, "show all X songs"). 

After this PR, the box width is only the content width plus some padding and it overlays also the dimmed texts. Also, the contents of the "New music available" box are now highlighted on hover.

Before:
before

After:
after

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -199,7 +199,20 @@
 	position: absolute;
 	top: 0;
 	background-color: #fff;
-	padding: 12px;
+	padding: 20px 50px;
+	z-index: 1;
+	width: auto;
+	margin-left: 50%;
+	transform: translateX(-50%);
+	-ms-transform: translateX(-50%);
+}
+
+#toScan:hover * {
+	color: #000;
+	/* opacity */
+	-ms-filter: "progid:DXImageTransform.Microsoft.Alpha(Opacity=100)";
+	filter: alpha(opacity = 100);
+	opacity: 1;
 }
 
 #updateData {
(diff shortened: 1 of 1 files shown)


MorrisJobke:
> When there are already some albums in the library, and new music files become available, the "New music available" box is overlayed on top of the album view. This box used to have quite odd layout as it spanned the whole screen width while its content could fit much smaller box. Also, it overlayed the album art and the track names but not anything written in dimmed font (track numbers, album years, "show all X songs").

oh. When I added this this was a little bar at the bottom of the screen and not that weirdly overlapping. Could we maybe go back to that layout again?

paulijar:
I don't remember seeing layout like that but I just started using Owncloud last summer. According to git history, @jancborchardt has changed the layout in commits afe62cc40bbfe and e9a76fd0cf88.

MorrisJobke:
> Without seeing the old layout, I could imagine that the new layout looks more balanced in the first-time-use case with empty library, which might have been the rationale behind the change.

👍

paulijar:
One more tweak: I made the background of the box just a tiny bit transparent.
transparency

gnarula:
Looks good to me :+1 Merged

Issue #518, opened 2016-07-26
Some embedded cover images are not shown


paulijar:
With the current development sources, Music app is able to show album cover images embedded in the MP3 files. This, however, does not work on all files. Especially, it never works if the image size is more than 16KB.

The root cause for this is the getID3 issue http://www.getid3.org/phpBB3/viewtopic.php?f=4&t=1930. A fix for this issue is now included in the main branch of getID3 (commit https://github.com/JamesHeinrich/getID3/commit/0e6c1a3bcef8219fede2c3ead13734296576bce5) but I'm not sure how the problem should be fixed for Music app.

Music app comes with its own version of getID3 but usually this version is not used. Instead, getID3 loaded by the core is used. On the other hand, getID3 has been already replaced with ID3Parser in the master branch of the core. ID3Parser is based on getID3 and has the same problem. I believe that core circumvents the problem by making a temporary copy of the music file before passing it to getID3 or ID3Parser.

gnarula:
Good catch! I've noticed this too and got the same error complaining about frame size in most of my files with large sized image covers. You're right, Music uses a separate copy of getID3 but core uses a minimal version of it (ID3Parser). Great that your fix has been merged in the main branch of getID3. I think we can upgrade the dependency once they publish a release with your patch applied :)

Another approach might be to use the Preview API in core (which does get around it by making a temporary copy) but that's not a part of the public API yet unfortunately.

gnarula:
@paulijar can you backport the changes to ID3Parser as well? I'd like to close this before pushing the next release and getID3 seems to have a long release cycle.

paulijar:
@gnarula Okay, I'll make a PR for ID3Parser when I have time. That might be tomorrow or next week.

paulijar:
@gnarula The fix is now included in ID3Parser release https://github.com/LukasReschke/ID3Parser/releases/tag/v0.0.3

paulijar:
@gnarula Are you planning to integrate the new ID3Parser or should I try to do it?

gnarula:
@paulijar I'll integrate the new ID3Parser, review the other PRs and build a new release in the next few days. Sorry for the delay, I've been a bit overloaded with exams and assignments in Uni lately.

paulijar:
@gnarula Okay, great. While testing my fix on ID3Parser, I noticed that it produces a bit different output array structure than getID3. So that's something to take care of while making the integration.

[...]

PR #547, opened 2016-10-23
Improve album art performance and compatibility

### LINKED ISSUE #534: Album art is not shown on old IE
The cover images or their placeholders are not shown on Internet Explorer versions 8 and 9. It seems to me that those old IE versions can't handle loading binary files with $http.get.

As a related issue, the cover image loading is still performing worse than it could. The existing cover images are still loaded twice: First in javascript $http.get is used to load the cover image. On success, the loaded data is thrown away and link to the image is added to the DOM tree. Then the browser loads the image again over network. Before #517, each picture was actually loaded three times and even missing pictures were tried to be loaded twice.

PRs #517 and #520 were crutches which made the cover image loading performance somewhat tolerable. But really the right thing to do would be to evaluate the availability of cover image already on the backend, and get rid of using $http.get in javascript.

paulijar:
After #517, loading the album cover art was still somewhat slow and did not work at all on IE8 and IE9. It was slow even in case none of the albums in the library had any cover art available; it took time before the placeholders for missing cover art got shown.

With this PR, the loading speed of cover images is improved by an order of magnitude. As a real-life example, my music library contains ~200 albums and almost all of them have embedded cover art set. Before the proposed change, hitting the reload button first took 10s before albums got shown, and then another 30s before all the cover images were loaded. After the change, it still takes 10s before the albums become visible, but then it takes less than 5s to show all the cover images. Placeholders for albums with no cover art are shown immediately.

The slowness and compatibility issues were caused by the fact that the cover images were loaded using javascript to run the HTTP GET queries. When modern browsers are allowed to do the loading directly from URL, they can perform much better. Javascript was used to run the queries because custom handling for failing queries was wanted: cover art was tried to be loaded for all albums and if the query failed, then the placeholder was shown.

The logic was now changed so that the availability of embedded cover art is checked at music scanning time, and the information is saved to the database. The database schema was not changed but the column cover_file_id in the oc_music_albums table is utilized in a new way: if cover for an album is found from a music file, then the file_id of that music file is stored in the column. Previously the column could point only to an image file residing in the same folder with the music tracks. The method ApiController.cover now checks if the pointed file is and image file or an audio file and extracts the image data accordingly.

Note that the music database has to be recreated after this change. If the database is not recreated, only placeholders will be shown for the albums with embedded cover art. It is probably best not to release this change before #518 is fixed, so that the users don't have to recreate the database twice.

Fixes #534.


paulijar:
It should also be noted that this PR changes the app behavior by making the Music app prefer embedded cover images instead of the distinct image files. Previously, embedded cover art was used only as a second choice if no image files were found from the directory of the audio file. 

This change was made to simplify the implementation, but I also think it makes more sense: picture file found from the directory of the audio file may or may not be related to the track in question, but the embedded cover art is definitely related to the track.

Issue #551, opened 2016-11-03
getID3 class loaded whether application is active or not


apotek:
(tried to use the template, but I get a 404 when I click on the link to it).

The way the getID3 class is loaded (in the music.php via `require`) causes class declaration conflicts with other applications that also use this library. 

```
// in stable5 getid3 is already loaded
if(!class_exists('getid3_exception')) {
	require_once __DIR__ . '/../3rdparty/getID3/getid3/getid3.php';
}
```

Can this be loaded via autoloader when the class is actually required, or, at least, can this be moved out of music.php, so that the class is not declared just by enabling the app whether it is currently active or not?

I am using audioplayer for music library management, and this app for being able to use ampache api to tomahawk. Both apps use getID3, which is a common dependency obviously for any multi-media app. 

Thinking it was an audioplayer issue, I created an issue for it [here](https://github.com/Rello/audioplayer/issues/78) https://github.com/Rello/audioplayer/issues/78, but in discussing with the developer, it became clear that the issue is not in that application's code, but in the way the getID3 class is being loaded in the music app.

Fixing the methodology of this class-load and removing it from music.php would increase interoperability with other applications.

Rello:
Hello,
I think the declaration needs to be moved into the `/music/utility/extractorgetid3.php`

because now we have the issue that music is claiming its local getID3 in a way that no other app can use its own.
This is especially difficult, as Music is distributing a pretty outdated version of getID3. Also other apps might need some local modifications of get ID3 which they can not distribute now.

can you please look at changing this?
thank you

apotek:
Here's a cheap PR for this. I know there's a better way, but I was worried about being more aggressive and breaking things. Sorry for the mistaken commits that I had to revert. https://github.com/owncloud/music/pull/552

Rello:
Thank you

apotek:
Previous PR was messy. This is one commit and with proper attribution https://github.com/owncloud/music/pull/553

It seems like travis doesn't have the mp3 test files uploaded :(

paulijar:
Travis has started failing like that somewhere between PRs #544 and #547, i.e. 2-3 weeks ago. I don't think that anything relevant has changed on Music app, so maybe Travis or its configurations have changed somehow. Interestingly, the tests run on CORE_BRANCHes stable8.2, stable9, and stable9.1 always pass and the others always fail.

Btw, Music will no longer distributes very old getID3 if PR #550 is merged.

Rello:
Hello, 
[...]

PR #552, opened 2016-11-03
Apotek get id3 inclusion

apotek:
This should address issue #551 . Loads getID3 library only when required by a process rather than during the load of the basic application. Allows for better interoperability between various OC media applications.


Rello:
That was quick
+1

apotek:
Git history here is dirty. I'll start over with a clean branch.

PR #553, opened 2016-11-03
Refs #551: Include get3ID libs as needed

### LINKED ISSUE #551: getID3 class loaded whether application is active or not

apotek:
Don't declare get3ID class at the basic app layer as it will conflict with other declarations simply by enabling the music app, whether in current use or not. Should fix #551

### DIFF
--- a/app/music.php
+++ b/app/music.php
@@ -45,11 +45,6 @@ use \OCA\Music\Utility\Helper;
 use \OCA\Music\Utility\Scanner;
 use OCP\AppFramework\IAppContainer;
 
-// in stable5 getid3 is already loaded
-if(!class_exists('getid3_exception')) {
-	require_once __DIR__ . '/../3rdparty/getID3/getid3/getid3.php';
-}
-
 class Music extends App {
 
 	public function __construct(array $urlParams=array()){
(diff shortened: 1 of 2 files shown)


Rello:
Hello,

these declarations also need to be moved (and rewritten partly?).
https://github.com/owncloud/music/blob/master/app/music.php#L256-L272

paulijar:
Merged with slight changes with 70fa45cbca4d3

PR #494, opened 2016-05-03
Oc9 compatibility

remiheens:
Add support for Owncloud Server 9

PR #266, opened 2014-08-10
[WIP] New playlist functionality (Final)

vgezer:
:construction: Work in progress!

Ready:
- [x] Sidebar with playlists
- [x] Hardcoded playlists
- [x] Hardcoded song display
- [x] Playlist - Database connection
- [x] Create basic playlist functionalities (create, edit, delete playlist)
- [x] Play the next song in the playlist after the current song
- [x] Add songs by drag&drop
- [x] Remove songs from playlists

Playlists:
- [x] Repeat playlist songs
- [x] Shuffle playlist songs

Working on:
- [ ] Design improvements

To do:
- [ ] Unit tests
- [ ] Documentation

Issues:

Apps menu is not shown: change z-index
Shuffle & repeat buttons are not shown, check css
Mobile view

Playlists may:
- [ ] be sorted by dragging

... and will be updated

PR #555, opened 2016-12-01
Another attempt to finalize the playlist functionality

paulijar:
This PR supersedes #266 and #494. I have rebased all the previous commits from these two PRs on top of the current master branch, made the changes necessary to operate on top of the new baseline, and added a few improvements.

Playlists can now be used but many improvements are still needed before this is good to release:
- [x] Allow dragging a whole album/artist to a playlist
- [x] Store the playlist track order in the database
- [x] Allow sorting tracks manually within a playlist
- [x] Indicate, which playlist is currently playing
- [x] Performance improvements
- [x] Special read-only playlist for "All tracks"
- [x] Layout tweaking
- [x] Show loading animation when moving from a playlist to the albums view

... and probably others which I cannot figure out right now.


gnarula:
Phew this has been one long PR indeed! Outstanding work as always @paulijar. A few trivial things that I noticed:

- When trying to delete the playlist that is currently active, perhaps we should switch to the Albums view?
- The "cancel" and "ok" icons move down on entering text in the New Playlist textbox. Please refer to this [image](https://www.dropbox.com/s/he5t66npbz71a3p/Screenshot%20from%202017-03-10%2020-35-50.png?dl=0)

paulijar:
@gnarula Great, thanks!

I agree with the deleting, moving automatically to the albums view would be a viable option.

I can't reproduce the layout issue seen on the screenshot neither on Firefox nor on Chrome. The layout seen in the screenshot is also otherwise a bit off. It might be that your browser hasn't reloaded all the css files properly. I have noticed that Chrome usually doesn't reload the css when you hit F5. You need to use CTRL+F5 (iirc).

gnarula:
@paulijar That's rather strange, I doubt it's caching issue. I was able to reproduce this while testing on IE11 on my friends laptop, who happened to use the app for the first time. Tried resetting the cache on chrome and Ctrl + F5 and it still persists on my end.

paulijar:
@gnarula Perhaps this is then a compatibility issue with different ownCloud/nextcloud versions. Some of the style definitions come from the core and if there are any changes, that might break the layout. I'm still using ownCloud 9.1.3.

PR #563, opened 2017-04-07
hand over ->fp for special WAV indexing

Rello:
the local fp needs to be used here because in certain cases, a temporary pointer is used for ID3v2 in WAV files

related to your:
https://github.com/owncloud/music/commit/330fc142a27d441808cdc370202718928407a2f4

also patched in Audio Player:
https://github.com/Rello/audioplayer/commit/6aeb7028226227d45b566e6ecb00cab110d1a741

cheers!


paulijar:
Makes sense. Merged. Thanks!

PR #565, opened 2017-05-05
Remove legacy call to vendor_script()

pellaeon:
It causes the php handler to run forever (thus a gateway timeout error) on rendering `templates/main.php`. The reason is unknown, but JS md5 is not needed in music anyway, I can't find anywhere in javascript that invokes md5.

Server version: Nextcloud 12 alpha
Database: sqlite


paulijar:
@pellaeon: Tested on ownCloud 9.1.3 + mysql and works fine. Btw, also the 'if' branch before the removed 'else' branch is obsolete since the Music app dropped support for ownCloud 8.0 and older on version 0.3.13. Would you like to remove that, too?

pellaeon:
Removed!

paulijar:
Thanks! Merged.

Issue #499, opened 2016-05-22
In Owncloud 9, shared files download, not play.


rolandixor:
They play fine if you are logged in, but the intention for this server is to have them be played in shared links.

rolandixor:
I hate to do this, but bump....

Seriously.

stratacast:
I will also bump

paulijar:
This issue is fixed in the brand new release v0.5.0.

Issue #464, opened 2016-02-02
Music "Light" request


traxxion:
This is a request for a version of the Music app that strips out all library functions and simply plays a file when clicked in the "Files" app and/or presents the files as in the "Files" app. 

The library functions of audio playback software are ever a hindrance to functionality (always have been) and assume that unlike any other files in the world, audio files somehow require ID3 tagging in order to be "organised". ID3 tags are extremely useful, but as a means of organisation are often mistaken.

If it is not a very difficult request, I would really appreciate a light version of Music for owncloud that removes the scanning and library database functions, or minimises them. This would also provide a fallback app for those who encounter library and db bugs. 

If it also presents the album art, either that already present in the folder or album art otherwise retrieved that would be good, as long as the mechanism is as simple as possible! 

Presenting ID3 tags read directly from the "current" file/folder would be nice.

A playlist function would be very beneficial.

Thanks

paulijar:
I, too, believe that if the user clicks on an audio file in the Files app, then more often than not she would just like to listen to that specific track. Throwing her to the Music app distracts the natural flow.

It is also possible that the clicked audio file is outside of the path which has been configured to be scanned by the Music app. In that case, the clicked audio file will not show up in the Music app and it is impossible to play it.

I would prefer to see opening an audio file from the Files app to work like this:
1. Clicking the audio file would bring up a small pop-up player and start playing the track.
2. The pop-up would have buttons for play/pause, a seek bar, and possibly a volume control.
3. The pop-up would also have a link "Show in music library" which would open the full Music app. This link could be hidden if the track is not within the path scanned by the Music app.

paulijar:
This feature is implemented in the brand new release v0.5.0.

Issue #433, opened 2015-09-28
can I stream music via shared link?


kiyoon:
I want to share what I listen or watch to my friends. Fortunately, I can share link of video and it supports streaming video for a link access too.

I want to stream music also, so I installed Music plugin but it did not solve the problem. They only can download the music but not stream and listen instantly.

I found that when I share the directory, it is able to stream music. But when I share just ONE music file, then it is impossible to stream. Sometimes I want to share just one music, not the entire directory.

How can I make my music streamable to all my friends who got the shared link?
I mean, shared link of only "one" music file.

Thank you.

rolandixor:
If I share a folder I still get this issue.

paulijar:
This seems to be a duplicate of the issue #256.

paulijar:
This issue is fixed in the brand new release v0.5.0.

Issue #256, opened 2014-07-29
allow publicly shared MP3s (via URL) to be streamed like videos do.


sahaq:
Is that a possibility? 

Thanks.

szepeviktor:
Yes. Shared MP3-s don't play in the browser. `$OC_Version = array(8,1,0,8);`

paulijar:
This issue is fixed in the brand new release v0.5.0.

PR #577, opened 2017-08-19
Small music player embedded into Files app

paulijar:
In this PR, a new embedded music player for Files app is created. When a music file is clicked in the Files app, the new player docks to the bottom of the screen and starts playing the file. The player has a play/pause button, seek bar, and volume control.

files-music-player

The new player works in all the views of the Files app (like 'All files', 'Favorites', 'Shared with others') and on publicly shared files and folders.

Before this PR, clicking a file in the Files app opened the full-blown Music app which kind of distracted the natural flow. Also, if the clicked file was not within the folder scanned by Music app, the file couldn't be played at all! Furthermore, this integration worked only in the 'All files' view. For public shares, on the other hand, an UI-less player was previously used but this worked only for shared folders and not for shared files.

This PR solves issues #464, #499, #433, and #256.


johnwry:
any updates on this option?

paulijar:
Exactly what kind of updates are you expecting on a feature which was completed and released six years ago?

PR #587, opened 2017-10-01
Fixed the buffer progress in Firefox.

Yurifag:
When playing a file in Firefox you will notice the buffer progress bar always stops at around 90-99%.
This is because the "progress" event doesn't fire when buffering has completed in Firefox.
Setting the buffer progress to 100% in the "suspend" event works around this issue as it only seems to be fired after buffering has been completed.
Note that the behavior in Chrome is a different one, where the "suspend" event is fired when buffering is suspended (_duh_).


paulijar:
This seems to work well. Great work, thanks!

For the record, the underlying issue in SoundManager2 is reported here: https://github.com/scottschiller/SoundManager2/issues/114.

paulijar:
With more testing, I actually found a case where this doesn't work quite correctly: if the file is large enough, also Firefox seems to suspend the buffering before the file has been completely loaded and emits the "suspended" event. I noticed this with a 46 MB .flac file.

PR #589, opened 2017-10-11
Implementation of downscaling images

### LINKED ISSUE #572: Downscale cover images on the back-end
This is a follow up task for #570.

It is possible that users have very large album cover images, especially when using an image file found from the folder of the track instead of an embedded cover image. The size of the image file may be e.g. several megabytes. At time being, the Music app loads all the cover images in full size from back-end to front-end where they are shown in relatively small size. If images are large, this causes lot of wasted bandwidth and time.

A better solution would be to downscale the image to, say, 200x200 pixels on the back-end and cache the downscaled versions. Size of such downscaled image should be around 10 - 20 KB.

Faldon:
Scales images on the back-end to a fixed with of 200px and a proportional height derived from the original source, if the dimension of the original image is bigger than 200px in either direction.

The downscaled image is stored in the cache as jpeg with 50% compression.

Closes #572

### DIFF
--- a/utility/coverhelper.php
+++ b/utility/coverhelper.php
@@ -17,6 +17,7 @@ use \OCA\Music\BusinessLayer\AlbumBusinessLayer;
 use \OCA\Music\Db\Cache;
 
 use \OCP\Files\Folder;
+use \OCP\Files\File;
 
 use \Doctrine\DBAL\Exception\UniqueConstraintViolationException;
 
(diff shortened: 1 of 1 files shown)


Faldon:
Hi.
I was a bit unsure about the compression factor, but you're right with 75% should be a good tradeoff.
And honestly, I didn't think of pictures with such odd dimensions. I'm rather picky with my cover images. -.-
I will move the logic to it's own function and also address the null pointer problem. I really didn't thought about this could happen in this branch of the code, but after re-reading I now see the need to move the function call.

I'll update the PR with your proposals later that day.

paulijar:
Looks great now. Cheers!

PR #590, opened 2017-10-15
Ampache: also accept authentication in "ssid" key

youknow0:
Until now, we only accepted the authentication token in the "auth"
key. This is the same as in Ampache.

However, for the URL of a song, Ampache passes the authentication
token in the "ssid" key. Some clients expect this key to have
this name, for example the Ampache Plugin for Rhythmbox, as shipped
with Ubuntu.

After this change, Rhythmbox works fine.

There is also a client-side fix for this problem:
<https://github.com/lotan/rhythmbox-ampache/commit/0d233e3282caea3f6a80c943bb414597a1afb733>

### DIFF
--- a/middleware/ampachemiddleware.php
+++ b/middleware/ampachemiddleware.php
@@ -62,8 +62,14 @@ class AmpacheMiddleware extends Middleware {
 		$this->isAmpacheCall = $annotationReader->hasAnnotation('AmpacheAPI');
 
 		// don't try to authenticate for the handshake request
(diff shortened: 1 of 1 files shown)


paulijar:
Okay, makes sense. Please follow the project style guide and indent with tabs instead of spaces.

paulijar:
@youknow0 I can merge this PR once you fix the indentations.

youknow0:
Sorry, I am a little busy right now. Indents should be fixed now.

paulijar:
Thanks!

Issue #598, opened 2017-12-01
Implement virtual scrolling 


Biont:
The current UI does not scale very well with large collections. This is because all artists and all albums are rendered to the DOM, no matter if the user is ever going to see them all. Plus, there are a number of event handlers on every one of them. This creates a lot of overhead the larger your music collection grows.

There are solutions to handle large lists effectively, like ionic collection repeat and [angular-vs-repeat](https://github.com/kamilkp/angular-vs-repeat).

Using these, the collection handler will ensure that only the visible segment of the list is actually being rendered, adding and removing elements as needed. This would make rendering performance in the music app a non-issue. (and you also get lazy-loaded cover art for free, since it won't be downloaded until you scroll it into view)

I have a working prototype using angular-vs-repeat in my fork and would love to PR it when it matures. (I hope to be able to push it soon, it's still a but ugly due to missing styling because of markup changes)

The only downside is that we would need to change the layout. For virtual scrolling to work, you need to know the exact height of the list (and all individual items). This is almost impossible with the current layout, since there can be 1 or dozens of albums for a single artist that reflow depending on the browser window width.

We need a fixed-height list. If elements can change size, it needs to be done with JS, so that we can have the collection manager calculate new offsets.

I could imagine having something similar to the collapsible track list next to an album: A fixed-size Artist with a fixed number of albums, and then a button that displays all of them if there are more.

What do you think?

paulijar:
The idea of virtual scrolling has been recorded for the first time in #78 some four years ago but so far it has not happened. I did some prestudy on the virtual scrolling libraries some half a year ago but came to the same conclusion as you: the current UI design of the Music app does not fit the virtual scrolling. As I didn't want to redesign the app, I then dropped the idea.

I kind of like the current UI design and how it adapts to different screen/window sizes. Also, in my opinion, the performance is nowadays quite sufficient on music collections with up to 10k tracks. The real problems start on collections with tens of thousands of tracks. But my gut feeling is that for the vast majority of the potential user base the 10k tracks is enough. So the question is, is it sensible to sacrifice even a little on the usability of the average user just to better serve the few heavy-duty users.

[...]

Issue #367, opened 2015-04-17
Ability to search through music


jancborchardt:
Similar to how it is in the Files app, Apps management etc, it should be possible to use the top bar search field to filter music.

As you type, the music list should be reduced. Everything should be searched over, including artist / album / song title / year / genre / etc. If that will result in performance hits, first just search by artist/album/songname.

boardfish:
Ctrl+F seems alright in the meantime, given things aren't lazy-loaded, but I'd appreciate this feature.

paulijar:
Search/filtering feature in included in the Music v0.13.0 which was just released.

PR #599, opened 2017-12-07
Virtual scrolling

Biont:
This is the pull request to implement virtual scrolling as discussed in #598

Instead of rendering the complete collection all at once, the changes in this PR will make the UI render only the visible segment of the collections and then add/remove elements as the user scrolls. This process should be unnoticeable for the user. It also makes switching views (e.g. "All Tracks" <-> "Albums" near-instant). This feature is implemented using [angular-vs-repeat](https://github.com/kamilkp/angular-vs-repeat) with a number of tweaks to make it work with our rather complex UI.

Here are a few details about the implementation to make it easier for you to review and understand what's changed:

### Breakpoints

The `albumGridService` provides the configuration object containing breakpoints and grid columns. It also has a few helper functions, most notably a method for calculating an artist's total height for all configured layouts. This enables us to precalculate all element sizes **once** during page load, instead of having to do it over and over again during layout changes or worse, while updating the list fragment.

### Responsive album grid

This is handled by the custom "vs-resize" directive which needs to be used on the same container as the vs-repeat (although it does not have a hard dependency on it. At least not yet). It is configured with the breakpoint object that the controller kindly injected into the scope for us.

It then does 3 things whenever a breakpoint is hit:  

 * Set the current layout as a data attribute
 *  change the "vsSize" property to the current layout, so that vs-repeat will use the new property for height calculations
 * trigger a re-render of the list

This solution turned out considerably less messy than I had expected it to be. It should even be possible to turn this into a general-purpose tool if we add more config options and handle CSS ourselves.

### Track lists

You will soon notice that I had to take some liberties with this one. And you can probably imagine that this part was one of the most troublesome to implement. Instead of pushing down the entire list, an expanded track list now opens a modal of sorts. It's still relative to the album and scrolls with it, but is dismissable by clicking outside of the list or scrolling it out of sight. It's also scrollable if the track list exceeds a certain (albeit generous) max-height. Why did I have to do this you ask?
 * Adding a dynamic offset to the list element and re-rendering would have added significant complexity to both the vs-repeat logic as well as the structure of the album grid.
 * You would have needed to keep track of the state (collapsed/expanded) of each album because the HTML is gone if scrolled out of view. (This is also why there is a max-height on the current solution)
 * Expanding the list *without* pushing the list down would overlap other albums and hide information. Therefore it needs to live on another layer.

While the existing UI allowed you to expand multiple albums at once, I did not feel this was especially useful: Unless the albums are directly next to each other, and you **really** only wanted to expand the albums in that particular row, that solution forced you to scroll and introduced immense amounts of whitespace, defeating the purpose of being able to see multiple expanded albums.

Even though you cannot have multiple lists open at the same time any more, I still find it more intuitive to quickly check out multiple albums: Expanding a list in the existing layout always caused a reflow, shoved elements around or pushed them out of sight. I like to think that the new popup reduces this kind of noise, since the underlying album list does not change any more and we now have a separate UI layer to show the track list.

### Alphabet Navigation

Another tricky part: Previously, this was done using du-smooth-scoll which allowed jumping to an anchor.
We don't have all the markup any more, so while it technically still worked, if would only be able to scroll within the segment that is currently visible. I created a new directive called "vs-scrollto" to handle scrolling.

It simply grabs the needed offset from the `sizesCumulative` array which vs-repeat luckily attaches to the scope for us. I know which offset to use by adding the corresponding artist's array-index to the lettersAvailable object and using it as an attribute in the navigation sidebar.

Again, this could easily be a general-purpose solution (which I honestly kind of expected to find in the first place), but it would require digging deeper into vs-repeat and how it interacts with its underlying ng-repeat. Pretty much the only thing that's missing is an easy way to access the complete collection from the outside and an attribute to configure which element property to compare against (--> so that we could jump to an artist by id, name, or whatever)


paulijar:
Thanks for the PR. I haven't really looked into the codes yet, but I did some testing on the UI. The best part of the proposed changes is that the application is now lightning fast to start and switch views. However, overall I feel that this PR is not ready to be merged. With a change like this, we should make sure we are not taking one step forward and two steps backward.

The most worrying aspect is that this change makes the scrolling feel less fluent. This is visible especially when dragging the scroll bar and when using the alphabet navigation. The effect is worse with larger music libraries, but already with 200 albums and 2000 tracks it's clearly visible. I'm afraid that this may be inherent to the architecture and not solvable. Or who knows, maybe there is something we can optimize to improve the feel.

I'm also not a huge fan of the track list popup thingie. From the point of view of usability, it's nearly always a bad idea to have a scrolling container within a scrolling container. To me, it also looks a bit weird how the track list shifts up and down when the popup is opened and closed. And on Firefox, there's an inactive scroll bar shown in the popup when the amount of tracks doesn't require scrolling.

Apart from the issues mentioned above, there are at least these bugs and problems which should be fixed before merging could be considered:

1. If the album has exactly 6 tracks, the track list is shown in a non-closable popup.

2. The track numbering does not work in the All tracks / Playlist view.

3. Clicking the song info area in the player controls bar is supposed to scroll the current album to view but now this doesn't work.

4. Similarly, when a file is played from the Files app and user clicks the song info area in the music controls bar, the file is supposed to be shown and played in the Music app. Now the automatic scrolling to the album in question does not work.

5. Scrolling with keyboard arrows or page up/down does not work properly. Sometimes it works for a while and then stops working.

6. Those layout breakpoints would need some adjusting. Now, when you start reducing the width of the window, the UI stays in the 3-column layout until track names mostly vanish, and in the 2-column layout until the track names vanish completely. And when the UI goes to 1-column mode, it happens at the same width where the UI enters the "mobile" mode with small cover art. There's never 1-column mode with large covers.

7. IMHO, the album covers should be in straight columns. I mean, if the UI uses 3-column layout, then all columns should be laid out according the 3-column spacing. Currently, if some row has only two albums, that row is laid out using only two columns and that makes the 2nd album look misplaced.

One more problem with this architecture is that it is not anymore possible to use the 'Find' feature of the browser to search within your music collection. The gap could be filled by implementing the search functionality within the app (#367). Looking at the code, there actually seems to be some kind of search integration, but it hasn't worked at least for the last 18 months while I've been involved with the project.

Biont:
>The most worrying aspect is that this change makes the scrolling feel less fluent.

Hm, could you elaborate this a bit more? I'm not sure I follow. There is the inherent lazy-loading of cover art (which should go away once the browser cache kicks in) which is very visible. Apart from that, guess it could only be the browser taking too long to render new fragments. There is the `vs-excess` attribute that adjusts how many elements to render ahead. Maybe this could by increased a bit to see if it helps

>I'm also not a huge fan of the track list popup thingie. 
Okay. I already noticed it looks like ass on FF. maybe there are ways to tweak the visuals, but I guess we'll have to find a different solution altogether. There are some other issues which I have hesitated to properly address because I feared I would be wasting my time. I'll see if I can make it expand the current row without adding too much complexity.

I'll try to fix the mentioned bugs soon. Thanks for pointing out the scrolling-related bugs. This is simply functionality I wasn't even aware of. I'm sure this is also the UI trying to jump to an anchor that's not on the DOM (or some variation of it), so the solution is probably very close to the way I made alphabet navigation work....hopefully.

>IMHO, the album covers should be in straight columns. 
I thought this was pretty cool since the current layout is very bad at dealing with long track names.
Might as well get rid of flexbox entirely then. This will also save us from that one IE8-guy that's bound to show up some day.

>One more problem with this architecture is that it is not anymore possible to use the 'Find' feature of the browser to search within your music collection.
Api support for this is pretty much nonexistent. Yeah, this would need to be alleviated by implementing the nextcloud search function inside the app.

Biont:
Speaking of implementing core features, how would you feel about moving the expanded track list to the right-hand side details pane that currently goes unused? We could freely scroll there and would not have to worry about the layout calculations at all. And on top of all, it would be completely in line with the rest of NC's ui

paulijar:
> >  The most worrying aspect is that this change makes the scrolling feel less fluent.
>
> Hm, could you elaborate this a bit more? I'm not sure I follow. There is the inherent lazy-loading of cover art (which should go away once the browser cache kicks in) which is very visible. Apart from that, guess it could only be the browser taking too long to render new fragments. There is the vs-excess attribute that adjusts how many elements to render ahead. Maybe this could by increased a bit to see if it helps

As long as you scroll slowly, the UI responds pretty well. But if you start dragging the "thumb" of the scrollbar and quickly scroll from the top of the albums list to the bottom, the UI looks like it gets refreshed like 5 frames per second (with 2000 tracks). Do the same with some 10k tracks and the UI refresh rate is maybe 1 frame per second. That was with Chrome. Doing the same on Firefox Quantum results in UI not being drawn at all during the scrolling, only empty white screen is shown.

I guess that this is caused because the browser is now adding and removing DOM elements at the same time as it should be animating the scrolling. In the old implementation, all the DOM work is executed first and then the UI becomes visible. Hence, there's less work to do during the scrolling. This of course happened on the expense of RAM usage which could become pretty dramatic with large music collections. Like most things in SW development, this is a trade-off. I'm not saying it's a bad one but just that I'm not 100% sure it's a good one either. Anyway, experimenting with `vs-excess` is probably a good idea.

> Okay. I already noticed it looks like ass on FF. maybe there are ways to tweak the visuals, but I guess we'll have to find a different solution altogether. There are some other issues which I have hesitated to properly address because I feared I would be wasting my time. I'll see if I can make it expand the current row without adding too much complexity.

Okay, good. I mean, also the popup solution has some benefits as you explained earlier. And I might be able to buy it with some tweaks. But still, I would prefer a solution which would be more similar to the current UI.

>>    IMHO, the album covers should be in straight columns.

>I thought this was pretty cool since the current layout is very bad at dealing with long track names.
Might as well get rid of flexbox entirely then. This will also save us from that one IE8-guy that's bound to show up some day.

I like it how the column width is adjusted to the screen width in your solution. I just don't like it that the number of columns varies row-by-row depending on the available content. The Music app has never really worked on IE8 so that's not a concern. IE9+, on the other hand, should work at least to some degree.

paulijar:
> The track numbering does not work in the All tracks / Playlist view.

There are also two more bugs which might be related to this one: 
- Clicking on any tracks of the All tracks / Playlist view starts playback of wrong song if the view is scrolled to anywhere else than the top of the list
- You should be able to reorder tracks of the playlist with drag-and-drop but now this doesn't work if the view has been scrolled from the top

paulijar:
> Speaking of implementing core features, how would you feel about moving the expanded track list to the right-hand side details pane that currently goes unused?

I wouldn't move the track list to the right-hand pane, I feel that it wouldn't be very good user experience when you expand a list and the expansion opens to another part of the screen than where you clicked.

I have actually been planning to add the details pane, though. My plan was that it could show track details which are not currently visible anywhere like file path, genre, comments, lyrics, etc. Using the same pane sometimes to list tracks and sometimes to show details of an individual track would be technically possible but maybe a bit confusing.

>  If - in the distant future - it will be possible to edit meta data, then this will be the right place to do so anyway

Even if/when we would be able to show all the metadata, I would be a bit hesitant to add support for editing it. The audio file tagging is quite complex issue with many different tagging formats and format versions. When writing the tags, there's a very real chance that we would be doing it in a way which is not compatible with the other tools/players the user uses. And I wouldn't want to be responsible of "breaking" the files of the users. As long as we are only reading the data, we cannot do catastrophic mistakes.

Biont:
I just grabbed Batarang and had a look at the performance issue (which I'd like to investigate first, since if we don't end up with significant performance improvements but actually make it worse, then what's the point of it all?)

During a regular page load (on the current release), I see more than 4 million watchers on average with a collection of about 2.000 songs. While this is probably absolute insanity given there's a recommendation to keep them under 2000 floating around, once the page is loaded, it's all more or less static and we only trigger new digests during playback, which does not require smooth rendering.

**With** virtual scrolling however, the underlying problem hits us full force. There just seems to be too much overhead to ensure smooth rendering when updating so often. The problem is that we have 2 levels of nested ng-repeats as well as quite a bit of logic in our template. Even though we're making heavy use of bindonce, ng-repeat will still introduce a ridiculous amount of watchers.

With some hasty hacks, I was able to cut the watchers down to a quarter, but consolidating that into something presentable will take a significant amount of refactoring.

Here's how I'd proceed:

1. Update angular to 1.3.* to get native one-time-binding which we can then use in a couple more places. There probably are some additional optimizations in the angular upgrade alone.

2. Turn the track-list (by far the worst offender) into some sort of self-contained widget. It will just receive some configuration and minimal markup to tell it how to render a single line. Then we just let it generate a static list without dynamic bindings (-->interpolate instead of compile) for individual tracks. Updating playback state and initiating playback can then be done at widget-level, with a single event handler/binding, instead of separate handlers for each and every track.

3. Push the changes and let you have a look. Until I get my actual music collection onto my dev system, I can only see the problem by adding additional ng-repeats. I suppose my test collection is simply nowhere near large enough to get problematic. (I wouldn't have pushed the PR if I had seen the problems before, after all)

4. If needed, try to do the same for the album grid. I guess getting rid of the ng-repeat for the track list will already net us some decent improvements since we're getting rid of an entire level of nesting (as well as the biggest one).

Once that's done, I will look at the scroll-position issues. In the meantime, maybe we can arrive at a consensus regarding the track list UI.


>Okay, good. I mean, also the popup solution has some benefits as you explained earlier. And I might be able to buy it with some tweaks. But still, I would prefer a solution which would be more similar to the current UI.

Can you imagine what those tweaks might be? **If** there's a way to make it work, then I'd prefer that over expanding the rows with unspeakable black magic. I'm unsure how to **not**
 make the popup scrollable since we cannot scroll the parent out of view. We'd need to completely detach it from the actual list so that it does not get killed along with the parent.

But even if we do that, the track list could be so large that it overlaps the album beneath it, making you unable to view its tracks entirely. Feels like we're going from bad to worse

paulijar:
> During a regular page load (on the current release), I see more than 4 million watchers on average with a collection of about 2.000 songs.

I didn't know of Batarang but now I also gave it a try. I believe you probably misread its output. With my collection of ~2000 tracks, I see that there are 9248 watchers in total and the average for last 30 digest cycles after loading is 2,803.033 watchers (note the decimal separator). That's still a lot but far from millions.

> Here's how I'd proceed...

Those sound like good ideas. If there are ways to significantly reduce the number of watchers without breaking anything, that's definitely a good thing.

You might want to consider splitting these improvements into several PRs. Many of them may be beneficial regardless of the virtual scrolling and doing them in separate PRs would ensure that all your hard work isn't wasted if we in the end can't find a good-enough solution for the virtual scrolling. And after all, that's how you eat an elephant: bit by bit.

> Update angular to 1.3.*

I once tested updating the Angularjs version but the new version couldn't be just dropped in. Apparently, the Music app uses some constructs which are not compatible with newer versions. I didn't dig any deeper, so it may be something simple or then not.

> Can you imagine what those tweaks might be?

1. The popup shouldn't start scrolling on normal-sized album, even if it is a multi-disk collection album. That means that the popup should be able stretch to some 60+ tracks. I'm okay with it scrolling after, say, 70 tracks. Maybe this could be achieved by increasing `vs-excess` so that the parent wouldn't go out of scope so soon?

2. The tracks should not move vertically when the popup is expanded, the first 5 tracks should stay on their exact positions. The popup opening animation should show how the list expands downwards.

3. The popup shouldn't look like ass on Firefox (in your words :D). Also the mobile layout would need some tuning, now there's too much of left margin.
As a separate but related issue, the albums now reserve the full height in the mobile layout even if there's less than 6 tracks. That unnecessary white space should be truncated.

Biont:
>(note the decimal separator)
Oh, thanks for pointing that out. :D Now it probably would have been easier to go from millions to a couple of thousands than it is to go from thousands to a few hundred. Damn, that means there's work do.

>You might want to consider splitting these improvements into several PRs.
Sure. My thoughts exactly. We just need to be careful not to create blockers.


Thanks for the hints on the collection popup thing. I'll see if it's doable.

omerths:
Hh

paulijar:
I think it's time to close this old PR. The track-list directive implemented in #614 already went a long way to improve the scalability. Now on Music v0.12.1, I have further built upon this by adding lazy-loading capabilities to the directive. This combination gives us most of the benefits of the virtual scrolling, while avoiding most of the downsides:
- Memory footprint is only slightly higher than in the solution proposed in this PR
- Scrolling remains smooth even with thousands of albums
- View switching is only slightly slower than in this PR (for All tracks view with 60k tracks it's significantly slower, because this view doesn't yet utilize the lazy-loading; this is on the work queue)
- Using browser search function to search by artist or album name works, but searching by track name is not possible. For this reason, integrating the ownCloud/Nextcloud core search function is my next goal.

Biont:
Yes I think this makes sense. Should you ever consider moving the frontend to vue (as there seems to be a push towards that framework in core) then actual virtual scrolling could be  thought through and solved at a fundamental level, together with a possible restructuring of the UI in general. The current 2-level view simply does not lend itself well to this feature. I would happily assist, by the way @paulijar

Issue #584, opened 2017-09-20
[Feature Request] ./occ music:scan --group option


mmattel:
Scan not only for a user / list of users, but also for a group like
``./occ music:scan --group=Family``
Would improve scanning resource needs :smile:

mmattel:
Update note:
Would need a logic to see which resources are common for the group and add the db entry for the particular file for many users. Anything what is not common is the classical way.

paulijar:
The way the index database of the Music app is built basically forces us to scan files separately for each user. So if there would --group option on the music:scan command, then it would be merely a convenience to save typing the names of the users belonging to the group. There wold be no performance benefit. As such, I don't think the the gains would justify the added complexity.

mmattel:
But eases admins life...
There is a files:scan --group option now in core, see https://github.com/owncloud/core/pull/30020

paulijar:
Okay, thanks for the info about `files:scan`. I guess that if you have large groups, then this might be a good feature: writing 3 user names on the command line is not much of a problem but writing 30 may be.

greku:
I'm starting to implement this small enhancement to get my hands on the music app. I think the description is quite clear to start with.

paulijar:
This feature has been implemented with the PR #613 and will be part of the next release.

PR #613, opened 2018-01-31
Implements --group option for the occ music commands owncloud/music#584

greku:
Here is the change for the --group option on the occ command.
It does not work exactly the same way as what was done on the files:scan --group.
files:scan --group explodes the group parameters with comma (to allow multiple groups), which doesn't seem to be right as owncloud allows group names with commas in it.

### DIFF
--- a/command/basecommand.php
+++ b/command/basecommand.php
@@ -24,9 +24,15 @@ abstract class BaseCommand extends Command {
 	 * @var \OCP\IUserManager $userManager
 	 */
 	protected $userManager;
+	/**
+	 * @var \OCP\IGroupManager $groupManager
+	 */
+	protected $groupManager;
 
-	public function __construct(\OCP\IUserManager $userManager) {
+	public function __construct(\OCP\IUserManager $userManager,
+			\OCP\IGroupManager $groupManager) {
 		$this->userManager = $userManager;
+		$this->groupManager = $groupManager;
 		parent::__construct();
 	}
 
@@ -43,6 +49,12 @@ abstract class BaseCommand extends Command {
 				InputOption::VALUE_NONE,
 				'target all known users'
 			)
+			->addOption(
+				'group',
+				null,
+				InputOption::VALUE_REQUIRED | InputOption::VALUE_IS_ARRAY,
+				'specify a targeted group to include all users of that group'
+			)
 		;
 		$this->doConfigure();
 	}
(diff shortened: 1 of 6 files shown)


greku:
The inspection indicates the code is "more complex". To me is acceptable, but I can try to clarify by adding a method, or rewrite differently.

paulijar:
Thanks for the PR. I'm rather busy at the moment, but I'll review it as soon as I have time.

command/basecommand.php
```diff
@@ -43,34 +49,69 @@ protected function configure() {
...
 			}
 			else {
-				foreach ($users as $user) {
-					if (!$this->userManager->userExists($user)) {
-						$output->writeln("User <error>$user</error> does not exist!");
-						$argsValid = false;
-					}
+				$users = $input->getArgument('user_id');
+				if (!$this->validateUsers($output, $users))
+					return;
```
paulijar: Please always use braces {} around the conditional blocks.

paulijar: Also, it's usually a good idea to avoid having multiple exit points in a function (unless the function is trivially simple). And four levels of nesting in one function is a bit too much for comfort. In this particular case, the cleanest solution would probably be to use exceptions. The function could be something like this:
```
protected function execute($input, $output) {
    try {
        self::ensureUsersGiven($input);
        $argUsers = $this->getArgumentUsers($input);
        $groupUsers = $this->getGroupArgumentUsers($input);
        $users = array_unique(array_merge($argUsers, $groupUsers));
        $this->doExecute($input, $output, $users);
    }
    catch (InvalidArgumentException $e) {
        $output->writeln($e->getMessage());
    }
}

protected static function ensureUsersGiven($input) {
    if (!$input->hasArgument('user_id')
        && !$input->hasOption('all')
        && !$input->hasOption('group')) {
        throw new InvalidArgumentException("Specify either the target user(s), --group or --all");
    }
}
```
The functions `getArgumentUsers` and `getGroupArgumentUsers` would extract the command line arguments and throw an exception if any of the arguments is invalid.

greku:
@paulijar Thanks for the hint about the exception. I agree it is the cleanest way to write the code.
Hopefully this time, Mr Scrutinizer will like it more.

paulijar:
Thanks, looks all good now. Good job!

PR #614, opened 2018-01-31
Custom track list directive

Biont:
## Description

We have touched this topic before in #599.

The current template for track lists involves a lot of logic and handles track display very dynamically, resulting in a large amount of watchers and CPU load during repaints. This pull request introduces a custom directive that renders and updates the track list with minimal overhead.

## Details
The list is now rendered only once (with as much vanilla JS as possible) and handles collapsing/expanding and track selection mostly via CSS. Click and drag events are implemented as a single delegated event handler, so we only need one per track list.

## Usage
The new markup boils down to this:
```html
<track-list
	click-handler="playTrack"
	more-text="'Show all {{ album.tracks.length }} songs …' | translate"
	less-text="'Show less …' | translate"
/>
```

The tracks, artist and album are then pulled from the scope, where 2 watchers are added.

## Benchmarks
I grabbed a few screenshots to illustrate the changes.
My test installation has a bit more than 500 songs. As each album/track list currently introduces more and more watchers, I expect the improvements to be even more drastic with larger collections.

#### Before:
pageload-master

#### After:
pageload-widget

The amount of watchers has halved as did the average digest time.
#### Before:
pageload2-master
#### After:
pageload2-widget

The whole music app is now loaded a full 2 seconds faster

#### Before:
playback-master

#### After:
playback-widget

This wasn't particularly slow earlier, but it's still nice to see such big improvements during playback as well.

## Things to watch out for.

- Drag&Drop could be a bit shaky. It worked wherever I tested it (-->adding tracks to playlists), but since the directive needs full control over its markup and tries its hardest to keep DOM manipulation and event handlers at a minimum, we cannot throw regular "ui-draggable" attributes on it (because that would require compiling the markup with angular and then spoiling most of the performance gains). The solution for this was to copy/reimplement what "ui-draggable" does to that it it is tricked into thinking it receives one of its own draggable elements.

- I did not fully understand how the l10n pipeline works (especially for pluralized strings). The more/less-texts might not work, but they don't seem to be translated right now anyway.

- It is currently not possible to define or alter the list item markup in the template. I initially planned to make the directive consume its innerHTML to use it as a template, but it did not feel right to leave it like that. Since the markup is only interpolated (-> variable replacement) instead of being fully compiled by angular, any of angular's custom directives and attributes would not work.

### DIFF
--- a/js/app/directives/tracklist.js
+++ b/js/app/directives/tracklist.js
@@ -1,0 +1,258 @@
+/**
+ * ownCloud - Music app
+ *
+ * @author Moritz Meißelbach
+ * @copyright 2017 Moritz Meißelbach <moritz@meisselba.ch>
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU AFFERO GENERAL PUBLIC LICENSE
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU AFFERO GENERAL PUBLIC LICENSE for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public
+ * License along with this library.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+
+/**
+ * This custom directive produces a self-contained track list widget that updates its list items according to the global playback state and user interaction.
+ * Handling this with markup alone would produce a large amount of watchers.
+ */
+
+
+angular.module('Music').directive('trackList', ['$window', '$rootScope', '$interpolate', function ($window, $rootScope, $interpolate) {
+
+	var tpl = '<div class="play-pause"></div>' +
+		'<span data-number="{{number}}" class="muted">{{number}}</span> ' +
+		'<span title="{{titleAtt}}">{{title}}</span>';
+
+	var trackRenderer = $interpolate(tpl);
+
+	return {
+		restrict: 'E',
+		link: function (scope, element, attrs) {
+			var rendered = false;
+			var expanded = false;
+			var listContainer;
+			var tracks = scope.album.tracks;
+			var moreText = scope.$eval(attrs.moreText);
+			var lessText = scope.$eval(attrs.lessText);
+
+			var listeners = [
+				$rootScope.$watch('currentTrack', function () {
+					requestAnimationFrame(render);
+				}),
+				$rootScope.$watch('playing', function () {
+					requestAnimationFrame(render);
+				})
+			];
+
+			/**
+			 * Render markup (once) and set classes according to current scope (always)
+			 */
+			function render () {
+				if (!rendered) {
+					var widget = document.createDocumentFragment(),
+						trackListFragment = renderTrackList();
+
+					listContainer = document.createElement('ul');
+					listContainer.className = 'track-list';
+
+					listContainer.appendChild(trackListFragment);
+					widget.appendChild(listContainer);
+					element.html(widget);
+					element.addClass('collapsed');
+					rendered = true;
+				}
+				/**
+				 * Set classes for the currently active list item
+				 */
+				var elems = listContainer.querySelectorAll(".playing, .current");
+				[].forEach.call(elems, function (el) {
+					el.classList.remove('current');
+					el.classList.remove('playing');
+				});
+				if (!scope.currentTrack) {
+					return;
+				}
+				var playing = listContainer.querySelector('[data-track-id="' + scope.currentTrack.id + '"]');
+				if (playing) {
+					playing.classList.add('current');
+					if ($rootScope.playing) {
+						playing.classList.add('playing');
+					} else {
+						playing.classList.remove('playing');
+					}
+				}
+			}
+
+			/**
+			 * Create the list of individual tracks. Skips after reaching the "toggle threshold"
+			 * so only tracks that are initially visible are actually being rendered
+			 *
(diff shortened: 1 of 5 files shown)


paulijar:
I haven't checked this thoroughly yet but here are some initial findings:
1. The latest commit doesn't include public/app.js, hence it doesn't work "out of the box"
2. Your branch should be rebased on top of the latest master branch
3. The placement of the "Show all X tracks..." is a bit off. It should be aligned on the same level with the track numbers.
4. The logic for deciding the amount of lines shown has changed. The logic should be like this:
```
IF album has 6 or less tracks THEN show them all
ELSE show 5 first tracks + the "Show all X tracks..." link
```
Hence, there's never more than 6 rows shown by default. This is the amount which fits nicely next to the album cover without visual "overflow".
5. If a track has no track number, then also the separator "." should be hidden
6. The tooltips have vanished from the tracks. The purpose of the tooltip is to show the full track name if it so long that it gets truncated. In case the album artist differs from the track artist, the tooltip contained also the name of the track artist.

Apart from these findings, this seemed to work well.

Biont:
Thank you for the feedback. I will make the needed changes soon

Biont:
@paulijar The screenshots in the OP are probably obsolete now that the PR containing the Angular update is merged. Logic dictates that there should still be a noticeable improvement, just probably not as drastic as shown above. I hope it's okay if I don't make new ones.

The changes are committed now. I also changed that the theme was wiring together the click event and the music playback. The directive now just emits an event which the overviewcontroller is listening to. This feels much cleaner to me and it reduces template complexity by another line.

js/app/directives/tracklist.js
```diff
@@ -0,0 +1,220 @@
...
+ */
+
+
+/**
+ * This custom directive produces a self-contained track list widget that updates its list items according to the global playback state and user interaction.
+ * Handling this with markup alone would produce a large amount of watchers.
+ */
+
+
+angular.module('Music').directive('trackList', ['$window', '$rootScope', '$interpolate', '$compile', function ($window, $rootScope, $interpolate) {
```
paulijar: Is '$compile' annotation unnecessary here?

Biont: Probably yes. I have been using this injection before and must have forgotten to remove it after deciding I won't need it. Will remove that tomorrow

paulijar:
Functionally, this seems to work OK now. Performance-wise, my results are two-fold: When Batarang is acitve, there is a huge difference in the performance for the benefit of the custom track list directive. Batarang is still slowing things down. But when Batarang is deactivated, it's hard to see any differences between the old and the new implementation. If there are differences, those are hidden behind the random variance.

Batarang shows that the amount of the watchers has dropped to ~1/3 of the original. But for some reason, this doesn't have any major impact on the perceived performance when running the application normally. Or at least I'm yet to find the use case where it would make a significant difference. I couldn't see any difference on memory footprint, either, although I didn't look at this too closely.

I tested this again on my test collection of ~600 albums. It might of course be that the real differences could be seen on even larger collection.

Biont:
@paulijar Could you describe your setup a bit? I'm honestly intrigued about it because on my machine, the difference is so massive that it's not even a comparison. 

Could it be that you have many albums with a very large amount of tracks in your test collection? Because that's the one disadvandage that this PR had over the current implementation before today: It rendered all tracks up front (albeit with display:none;) while the angular template dynamically skipped rendering the additional markup if the track list was collapsed. It **could be** that given a immensely high track/album ratio, the benefits of this new directive were overridden by the additonal markup (and processing thereof)

In my last commit, I got rid of this disadvantage. Now only visible tracks are rendered. Upon expanding a track list, they are lazy-loaded and inserted.

To prove my case, I went ahead and got some numbers. First I wanted to get the DevTools out of the equation (not just Batarang) to demonstrate how much faster a full page load becomes:

```
* No DevTools open
* Stop watch is started while pressing F5 and stopped as soon as albums are popping up (ignoring the loading of album covers)

master:
3.43
8.88
4.03
5.99
6.12
6.66
3.87
6.10
5.97
3.74
----
avg: 6.02

track_list_widget:
2.05
1.93
1.67
1.52
1.59
1.67
1.81
1.85
1.65
1.79
----
avg: 1.75


Difference: 70.93%

```

Then I wanted to get closer to where the optimization actually takes place: When rendering the album list itself. To get a better look, I used the CPU throttling feature of the DevTools to artificially slow down my machine and give me more time to react.

```

* DevTools open
* Batarang disabled
* CPU-throttling 4x
* Stop watch is started once the sidebar is rendered (->after "Scanning music" has completed) and the loading circle appears 
* Stop watch is stopped once the album list pops up (ignoring the loading of album covers)


master:
24.13
24.05
24.53
22.69
25.63
25.06
26.40
24.96
26.58
27.03
----
avg: 25.10

track_list_widget:
4.18
4.00
4.18
4.25
3.37
3.76
3.86
3.47
4.11
3.88
----
avg: 3.90

Difference: 84.46%

```

Please check again with my last commit. Maybe your test setup needed this last boost.

paulijar:
Okay, now I can see a significant improvement. And you are right, my "large test" library has somewhat unrealistic distribution of 58625 tracks in only 558 albums. A few dozens of the albums are real-life albums with 2-20 tracks but the bulk of the library consists of factored test albums with 112 tracks each.

The other library I was testing was my real-life collection of 178 albums and 1954 tracks. This library is still so small that the rendering is almost a non-issue even on the master. Hence, it was difficult to see any differences when measuring by hand. But my gut feeling is that also this got a bit faster with your latest optimization. And that does makes sense, with 10+ tracks per album not rendering the hidden ones saves ~50% of the html elements.

Biont:
Hmm... I wanted to rebase my fork so that you have a singular commit to merge instead of my ~30 chaotic WIP commits, but while everything looks ok in my branch, the commits still show up above. The links are now dead, so maybe it's just displayed incorrectly.

PR #615, opened 2018-02-01
Update AngularJS to 1.6.8

Biont:
## About

Our version of AngularJS is currently still at 1.2.21 from 2014. The framework has seen a lot of development in the meantime and we could benefit from a boatload of bugfixes, enhancements and performance improvements simply by getting this library up-to-date.

One of the main benefits would be the introduction of native one-time-bindings in 1.3 which could be used in more ways than the bindonce-library that we employ for this purpose.

So my initial goal was to bring us to 1.3 at least, but it turned out we can go all the way to the current release.

## Procedure

I tried to be very careful when updating the library, assuming that just throwing in the newest version would result in hundreds of errors that I would never be able to fix all at once. So I went through almost every single release and checked for errors after the update. This is what I did:

1. Update to new version
2. Reload music app
3. Start playback
4. Navigate to the track list view and back
5. Use the alphabet navigation
6. Commit the update if no issues were found

I'll just hope that any glaring regression would have shown up during this test. There were a few minor issues that I fixed, but nothing huge. I commented on such issues in my commit messages.

## Results

To my surprise, the newer versions perform signicantly better than the old one. See the attached screenshots for comparison.

#### 1.2.21
pageload-master

#### 1.6.8
pageload-angular-new

#### 1.2.21
pageload2-master

#### 1.6.8
pageload2-angular-new

#### 1.2.21
playback-master

#### 1.6.8
playback-angular-new

Note that this is without any changes to templates or logic other that the bugfixes needed to get the new version working. In a follow-up PR, I would like to implement the new one-time-bindings for even more improvements.


paulijar:
Great, thanks! I'm rather busy right now but I'll take a look as soon as I have time.

paulijar:
I gave it a try, and as far as I can tell, everything still works. Tested Firefox, Chrome, and Edge.

It's really hard to actually see any performance improvement on daily use. However, with a stop watch I could see that on a (somewhat unrealistically distributed) test library of 558 albums and 58625 tracks the opening time has improved from ~10s to ~8s. Measured on Chrome by navigating from the Files app to the Music app and stopping the clock when the albums become visible.

Good job!

Issue #620, opened 2018-02-12
Building js with make causes error


greku:
In build folder, running make causes this error:
![screenshot_2018-02-12_22-14-45](https://user-images.githubusercontent.com/259454/36120091-57d08ff2-1042-11e8-9bee-dbf1d8375763.png)

It seems to be due to the dest parameter set to empty string in the Makefile.

PR #621, opened 2018-02-12
gruntfile.js fixing wrap dest parameter causing error #620

greku:


paulijar:
The issue is not reproducible on my installation but the fix doesn't seem to cause any problems, either. Hence, it's good to merge.

PR #622, opened 2018-02-13
Switch to native bindonce syntax, remove bindonce-library in the process

Biont:
In #615 we updated AngularJS to the most recent version. This allows us to use the "new" native one-time binding functionality that used to be implemented with a helper library.

The new syntax is a lot cleaner. Instead of having to use `bindonce` and `bo-*` attributes everywhere, we can simply prefix variables with `::` which I did throughout the template files.

With the new syntax, we can also use `ng-repeat` more efficiently, since -together with the `track by`-we can now trim down the amount of needed watchers considerably. I don't have detailed benchmark results this time because I consider this PR to be more of a maintenance&cleanup than a performance thing, but I went from ~1130 to ~430 in my installation.

So we have

- cleaner markup
- one less 3rd party lib
- measurable (~~not necessarily _noticeable_) performance improvement.


paulijar:
Yes, this makes sense but there are now at least a couple of bugs.

1. On my system, the application main view didn't show any albums when I pulled these changes. AngularJS gave an error like this:
```
Duplicates in a repeater are not allowed. Use 'track by' expression to specify unique keys. Repeater: artist in ::artists track by artist.id | limitTo: incrementalLoadLimit, Duplicate key: 3718, Duplicate value: {"id":371817,"name":"Apocalyptica","albums":[{"name":"Plays Metallica by Four Cellos","year":"1996","disk":1,"cover":"/nextcloud/index.php/apps/music/api/album/197948/cover","id":197948,"tracks":[{"title":"Enter Sandman","number":1,"artistName":"Apocalyptica","artistId":371817,"files":{"audio/mpeg":43006},"id":339106},{"title":"The Unforgiven","number":4,"artistName":"Apocalyptica","artistId":371817,"files":{"audio/mpeg":43086},"id":339126}]},{"name":"Inquisition Symphony","year":"1998","disk":1,"cover":"/nextcloud/index.php/apps/music/api/album/197961/cover","id":197961,"tracks":[{"title":"Nothing Else Matters","number":4,"artistName":"Apocalyptica","artistId":371817,"files":{"audio/mpeg":43058},"id":339119},{"title":"One","number":11,"artistName":"Apocalyptica","artistId":371817,"files":{"audio/mpeg":43062},"id":339120}]},{"name":"test tracks %% %20","year":"2000","disk":1,"cover":"/nextcloud/index.php/apps/music/api/album/197949/cover","id":197949,"tracks":[{"title":"Fight Fire with Fire (live)","number":null,"artistName":"Apocalyptica","artistId":371817,"files":{"audio/mpeg":43010},"id":339107},{"title":"In Memoriam (live)","number":null,"artistName":"Apocalyptica","artistId":371817,"files":{"audio/mpeg":43038},"id":339114}]},{"name":"Cult Special Edition","year":"2001","disk":1,"cover":"/nextcloud/index.php/apps/music/api/album/197946/cover","id":197946,"tracks":[{"title":"Path","number":1,"artistName":"Apocalyptica","artistId":371817,"files":{"audio/mpeg":43070},"id":339122},{"title":"Struggle","number":2,"artistName":"Apocalyptica","artistId":371817,"files":{"audio/mpeg":43082},"id":339125},{"title":"Romance","number":3,"artistName":"Apocalyptica","artistId":371817,"files":{"audio/mpeg":43078},"id":339124},{"title":"Pray!","number":4,"artistName":"Apocalyptica","artistId":371817,"files":{"audio/mpeg":43074},"id":339123},{"title":"In Memoriam","number":5,"artistName":"Apocalyptica","artistId":371817,"files":{"audio/mpeg":43042},"id":339115},{"title":"Hyperventilation","number":6,"artistName":"Apocalyptica","artistId":371817,"files":{"audio/mpeg":43034},"id":339113},{"title":"Beyond Time","number":7,"artistName":"Apocalyptica","artistId":371817,"files":{"audio/mpeg":42997},"id":339104},{"title":"Hope","number":8,"artistName":"Apocalyptica","artistId":371817,"files":{"audio/mpeg":43030},"id":339112},{"title":"Kaamos","number":9,"artistName":"Apocalyptica","artistId":371817,"files":{"audio/mpeg":43050},"id":339117},{"title":"Coma","number":10,"artistName":"Apocalyptica","artistId":371817,"files":{"audio/mpeg":43001},"id":339105},{"title":"Hall Of The Mountain King","number":11,"artistName":"Apocalyptica","artistId":371817,"files":{"audio/mpeg":43018},"id":339109},{"title":"Until It Sleeps","number":12,"artistName":"Apocalyptica","artistId":371817,"files":{"audio/mpeg":43090},"id":339127},{"title":"Fight Fire With Fire","number":13,"artistName":"Apocalyptica","artistId":371817,"files":{"audio/mpeg":43014},"id":339108},{"title":"Path vol. 2","number":14,"artistName":"Apocalyptica feat. Sandra Nasic","artistId":371851,"files":{"audio/mpeg":43066},"id":339121},{"title":"Hope vol. 2","number":15,"artistName":"Apocalyptica feat. Matthias Sayer","artistId":371831,"files":{"audio/mpeg":43026},"id":339111},{"title":"Harmageddon (live)","number":16,"artistName":"Apocalyptica","artistId":371817,"files":{"audio/mpeg":43022},"id":339110},{"title":"Nothing Else Matters (live)","number":17,"artistName":"Apocalyptica","artistId":371817,"files":{"audio/mpeg":43054},"id":339118},{"title":"Inquisition Symphony (live)","number":18,"artistName":"Apocalyptica","artistId":371817,"files":{"audio/mpeg":43046},"id":339116}]}],"route":"collection","reqParams":null,"restangularized":true,"fromServer":true,"parentResource":null,"restangularCollection":false}
```
The `track by artist.id` and `limitTo:` declarations now seem to interact badly. Note how angularjs seems to be using only the four first digits of the six-digit artist.id. 

This got fixed when I removed the `| limitTo: incrementalLoadLimit` filter but I'm not sure if we can do this as a proper fix. The purpose of the incrementalLoadLimit is to keep the browser responsive when loading the view with a huge collection. Your track list directive PR probably made this much less important, but it might still be necessary with large enough collection especially on slower browsers. This would require some testing.

2. It should be possible to reorder the custom playlists by drag-and-drop and to remove tracks from the list by clicking the X-button after the track name. These don't now work as you are using one-time-binding for the playlist contents.

Biont:
I'll have a look at that. Thanks for the feedback.

Biont:
@paulijar Could you have a look again? Sorry for the delay, real life got in the way hard.
Anyway, I read that the track by is supposed to go after the limitTo, but I had no problems on my setup in either case. If it still causes problems, I would maybe just remove either the track by or the limitTo :)

Track list reordering now works again

paulijar:
Thanks for the update.

I still get the error in the albums view. Instead of the element `artist-area` causing my problem, you had modified the `ng-repeat` in the `album-area` element by adding the `limitTo` block. Was this by mistake? Because to me, this modification doesn't seem to make much sense.

Anyway, I tested changing the order of `limitTo` and `track by` in the `album-area` element but this didn't work correctly, either. It did solve my earlier problem and the albums view now showed up but it was permanently limited to four first artists.

Also the playlist view still has some issues: when the list is modified, the ordinal of the playlist track is not updated accordingly. The numbering is updated only if the user navigates away and back to the list. Another, quite weird problem is that whenever you drag and drop a track on a playlist, the highlighting of the active view disappears from the navigation pane.

paulijar:
The Travis failure is not caused by your changes. The tests on ownCloud have been failing for some time now. Tests on Nextcloud work fine. I don't know the reason but it doesn't seem to be any change in the Music app which causes these failures: also such branches which passed earlier are now failing systematically.

Biont:
Alright let's just remove the fancy collection modifiers and concentrate on the new syntax here. I will push the changes soon

Biont:
@paulijar Reordering now properly updates the index. I have removed the 'track by' until I can reliably reproduce the issues you were having. 

>Another, quite weird problem is that whenever you drag and drop a track on a playlist, the highlighting of the active view disappears from the navigation pane.

This is something I either fail to understand or cannot reproduce.

1. I'm in "All Tracks"
2. I drag&drop a song into one of my playlists
3. The sidebar stays the same.

If I understood correctly, you were saying that "All Tracks" would no longer be selected after dropping the track?

paulijar:
This still doesn't work correctly:
- The Albums view shows albums only for the first 4 artists. Angularjs is apparently evaluating the whole statement `artists | limitTo: incrementalLoadLimit` only once.
- Newly created playlists don't appear to the left pane. We shouldn't one-time-bind the set of playlists, either.

> This is something I either fail to understand or cannot reproduce...

You understood basically correctly. However, this doesn't seem to happen if you drag a track from the *Albums* of *All trakcs* views, you need to drag a track from one of the custom playlists. It doesn't matter if you are dragging the track to reorder the list or to copy the track to another playlist.

Biont:
We'll get there eventually :blush:

paulijar:
As this PR has been stalled for quite some time, I took the liberty of implementing the change in commit 
1ecc855. I did only straight-forward conversion from the `bindonce` library to the `::` syntax without trying to make any further optimizations. Hence, this PR is obsolete.

Btw, I realized that the problem of disappearing navi-pane highlighting after dragging a track from playlist had nothing to do with the one-time-binding changes. The bug has probably been there since the introduction of the navigation pane and playlists but I just hadn't noticed earlier. Now it's fixed by 0352bbd.

Biont:
Sorry for not progressing here for so long.

paulijar:
> Sorry for not progressing here for so long.

No problem

PR #625, opened 2018-02-23
Settings mainview

greku:
Here are the changes to access the settings of the music app from the main app. The setting form layout remains the same as before, and the behaviour also very similar.


paulijar:
Thanks for the PR. The codes look good. However, I noticed a few issues on testing: 

1. You are missing `use \OCP\IURLGenerator;` in settingscontroller.php. On my system, this made the REST API return code 500 and the settings view did not open at all. Maybe this depends on the PHP/server configuration if you haven't noticed it.

2. If there are many playlists or the window height is very low, a scroll bar is needed for the left pane. Currently, the scroll bar eventually appears but it does not work correctly. It is impossible to scroll the pane all the way to the bottom to see the last playlists.

3. The highlighting of the Settings item in the left pane should work the same way as with the other left pane items: on hover and when the view is open. Currently, the highlighting seems to work correctly on ownCloud 9 but on Nextcloud 12 the Settings item is not highlighted on either case. On Nextcloud master branch (14.0.0-alpha) the highlighting works on hover but not when the view is open. I didn't test other cloud versions. I think that NC12 core may have some bug here because on NC12 the highlighting of the Settings item doesn't work in the Files app, either.

4. This is very minor, but the settings link in the left pane now works differently on narrow window ("mobile layout") than the other links. On all other view changing links, clicking the link also collapses the left pane but on the settings link this doesn't happen.

5. When the path to music collection is changed, the app should also wipe clean the tracks and albums loaded from the old path. Now the app keeps showing the tracks which are no longer in the database until when the rescanning is started. The situation is especially odd if the user would have chosen a folder which does not contain any audio files: The "No music found" label is shown over the previously loaded albums.

6. When the scanning finishes, the app automatically reloads the music collection. If the settings view is open when this happens, the view doesn't reload correctly and the loading spinner is shown indefinitely.

PS. The Travis CI failure is not related to changes of this PR. The same failure now happens on the master branch head contents while the same commit previously passed the test. I guess there's something wrong on Travis itself.

greku:
Thank you for the review and detailed comments. I will have a look to those issues.

greku:
I think now the 6 points that you mentioned are solved.

greku:
I installed NC13 to do more tests and see a problem with the settings link style.

paulijar:
Thanks for the update.

In addition to the link style problem, there's now problem also with the positioning of the playback controls on NC13. I believe that your changes are not compatible with my recent fix  38a4e2a8eb. This commit of mine may actually be just unnecessary after your changes.

Also, the scroll bar and the highlighting still don't work correctly on Nextcloud 12. On ownCloud, those work nicely (tested 9.1 and 8.2).

There's now also one new issue: visibility of the loading indicator does not work properly in the playlist view. When the view is loaded, the indicator is hidden before the content is actually visible. For a short while, there's only blank screen with no content and no loading indicator. You need more than a handful of tracks to see this but at least on collection of ~2000 tracks it was clearly visible when opening the "All tracks" list.

greku:
I tried it on my owncloud setup with 700 tracks and firefox/chrome, but I can't see any effect of switching many times the view between All tracks and Settings. I also looked at the developer tool from chrome but don't see any abnormal increase of memory usage. It seems the garbage collector recovers memory correctly after a few switches.
It must be related to the larger collection that you have. Don't you see this problem when switching between All tracks and Albums? 
I suspect I will need help to solve this, I don't really have an idea.

paulijar:
After some more testing, I found that you indeed can reproduce the problem also by navigating between All tracks and Albums. Furthermore, it can be reproduced also without changes from this PR. It's just that the problem becomes visible with much smaller amount of navigation steps when switching between All tracks and Settings views. On the other hand, the problem is not present when navigating between Albums and Settings views.

So probably the root cause is now in the All tracks view. Something in the activation of the Settings view just amplifies the adverse effect.

greku:
I have now over 1800 tracks, and I can confirm that it loads more slowly than before. I think the fact it takes time to load is just because there are many tracks. Time to load/unload the view is around 2-3 seconds with firefox (52.6.0 (64-bit)) on desktop PC. Chrome is much faster, less than a second.

The difference I see between the All tracks and Albums, is that All tracks view is deactivated by calling show_less incrementally in steps of 1000, where as the albums is in steps of 4.
On my setup, when leaving the all tracks view, it takes a bit of time before the loading icon appears. It is more instantaneous if I change the incremental load step to 10 instead of 1000.

Still, I don't have an obvious effect that it loads very slowly when switching a few times the view.
So, if you have time, one test that could be interesting is that you would change the file playlistviewcontroller.js
`var INCREMENTAL_LOAD_STEP = 1000;`
into something else like:
`var INCREMENTAL_LOAD_STEP = 10;`
and tell me whether it makes any difference?

paulijar:
I'll do some more testing when I have time. 

Note that in the albums view the increment is 4 *artists* while in the playlist it is 1000 *tracks*. And after #614, the "optimal" increment for the albums view would probably be something greater than 4 (it's anyway just a simple heuristics and music collections differ so no step size implemented like this could be truly optimal for all users).

Biont:
I added some (jqLite) event listeners in #614 that I do not properly cleanup in '$destroy', either assuming they'd be gone after deleting the node, or simply forgetting about them. I will revisit this code, maybe it's the cause of the leaks, even though I wasn't able to reproduce a performance issue in a quick test

paulijar:
I created a new issue #635 for the "navigation slowing down" problem as it's an existing issue and we don't need to block this PR because of it.

There's one more thing which I think would need some revising, and that's the use of the `loading` and `loadingCollection` flags. At the moment, that's quite confusing and it's difficult to tell what each of these actually means. The sensible way would be that `loadingCollection` would only mean that fetching the `api/collection` is ongoing, and `loading` would mean that view switching is ongoing. To achieve this, `loadingCollection` should be modified only from maincontroller.js.

I did some testing on the flags and I think it works nicely with these changes:
1. Rollback changes of commit b85becf05e on these files: overviewcontroller.js, playlistviewcontroller.js, sidebarcontroller.js.
2. Add `$rootScope.loadingCollection = false;` to the line 90 of maincontroller.js, to the success-callback of ArtistFactory.getArtists().

greku:
Thanks again for the review. I agree it is more clear as you propose to set the loadingCollection. I changed it in the way you propose and don't see any problem with it.

paulijar:
Thanks. Looks good now. This turned out to be rather large task but now I'm happy to merge it. Well done!

paulijar:
@greku I forgot to answer your question earlier, but setting `var INCREMENTAL_LOAD_STEP = 10;` did not have any effect on the *slowing down* issue. It just made the view switching slower overall (because 100 times more asynchronous steps were now needed to show/hide the contents), but it still got even slower with repeated view switches.

greku:
@paulijar Thanks for trying anyway and also for merging the PR. The issue itself is something that seems to be very dependent on the web browser potentially the rest of the setup. So far I couldn't reproduce it on my setup. Firefox is definitely slower than Chrome on this, but I can't see the increased slowness when switching views again and again. Hopefully, Biont will find out while he is busy with similar performance aspects.

paulijar:
This feature is now published as part of release v0.6.0.

Issue #626, opened 2018-02-24
Usability issue with "+ New Playlist" on the mobile layout


paulijar:
When the browser window is very narrow, the "mobile" layout is used. Here, the left pane is hidden by default and shown with a toggle button.

There's now a usability issue when the user clicks the link "+ New Playlist". This auto-collapses the pane just when the user is supposed to write the name for the new playlist. Auto-collapsing makes sense when the user presses a view-switching link like an existing playlist or "Albums" but not here.

paulijar:
The fix is published in v0.6.0.

PR #628, opened 2018-02-25
fixes owncloud/music#626 pane closing on mobile when creating new pla…

greku:
The snapper is part of the core, the code foresees a class to avoid collapsing if not needed. I added this class to the links and buttons involved in the playlist creation process.


paulijar:
This does the trick, thanks!

PR #637, opened 2018-03-21
Do no use subselect in updateFolderCover method, for performance reasons

jmdeboer-surfsara:
We are using OwnCloud in a large installation (> 30.000 users). After enabling the music app, we started noticing a gradual increase in database server load.

3m

We eventually tracked it down to this query:
UPDATE `oc_music_albums` SET `cover_file_id` = ? WHERE `cover_file_id` IS NULL AND `id` IN ( SELECT DISTINCT `tracks`.`album_id` FROM `oc_music_tracks` `tracks` JOIN `oc_filecache` `files` ON `tracks`.`file_id` = `files`.`fileid` WHERE `files`.`parent` = ? )

Which in some cases took multiple seconds to complete (on beefy hardware). The subselect within yielded 0 rows so you'd expect the complete query to be quick, but the problem is that the outer query yielded a lot of rows (over 20.000), so we ran into this limitation of MySQL:

https://dev.mysql.com/doc/mysql-reslimits-excerpt/5.5/en/subquery-restrictions.html

"A typical case for poor IN subquery performance is when the subquery returns a small number of rows but the outer query returns a large number of rows to be compared to the subquery result."

By executing the two queries separately, the MySQL optimizer works a lot better and the server load dropped instantly.

screenshot_2018-03-20_15-16-09

### DIFF
--- a/db/albummapper.php
+++ b/db/albummapper.php
@@ -155,19 +155,25 @@ class AlbumMapper extends BaseMapper {
 	 * @return true if one or more albums were influenced
 	 */
 	public function updateFolderCover($coverFileId, $folderId){
-		$sql = 'UPDATE `*PREFIX*music_albums`
-				SET `cover_file_id` = ?
-				WHERE `cover_file_id` IS NULL AND `id` IN (
-					SELECT DISTINCT `tracks`.`album_id`
+		$sql = 'SELECT DISTINCT `tracks`.`album_id`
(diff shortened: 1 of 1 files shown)


paulijar:
Thanks for the PR. This makes sense, and according to brief testing, the function still does its job.

paulijar:
This feature is now published as part of release v0.6.0.

PR #640, opened 2018-04-26
[Experimental] Add stub implementations for core APIs to be used by Scrutinizer

paulijar:


paulijar:
Hmm, this doesn't seem to work as intended. Discarding.

PR #649, opened 2018-06-09
music logotype horizontal

nunojesus:
Here are the PNG and SVG files of the logo, I send also the 1000px logo as you requested.

The font of the logotype is the: Righteous

If you need something more just say it, Im here to help.

Greetings!

Nuno Jesus


paulijar:
Great, thanks! In the end, I decided to keep only the svg versions of the files as the png versions took some space and were not needed for anything for now. It's easy enough to restore them from git or recreate from the svg if later a need arises.

I also added the logotype to the top of the github landing page.

PR #658, opened 2018-08-07
Border Radius   ???? % ???

blackcrack:
since when it is give Border Radius in percent ? it is ever px imho :)

best regards
Blacky


paulijar:
5 seconds of googling gave me this: https://www.w3schools.com/cssref/css3_pr_border-radius.asp

According to this, percents is a perfectly valid way to define border radius.

At the moment, the volume slider looks like this on Firefox: 
image

With your change, it would look like this:
image

What is a bit silly is that the radius has been defined as 100% because 50% is the maximum and makes the element a perfect circle. Defining anything above 50% doesn't change the result from this at all.

blackcrack:
@paulijar so should we setting up 50% ?
best 
Blacky

paulijar:
Well, setting the radius to 50px has obviously the same result as setting it to >= 50% because the size of the slider thumb is maybe ~20px. Also setting it to anything > ~10px would have the same effect.

But I don't really see your point here. Why would we make this change which actually doesn't change anything? If anything, it makes the style more difficult to interpret: setting the radius to >=50% is guaranteed to always produce a perfect circle and that's what we want. When the radius is set in pixels, the shape depends on the size of the element which is being rounded.

blackcrack:
it was setting up a radius with 100% from anyone who have no clue, what ever,
so if we change it, then right.. so, let it define as round corners and not as scale round via %
this is why i have now setting up as round with 50px who was before 100% (what was slightly exaggerated \*g*)
so this was my intension, "make it right.."

best regards my Frend :)
Blacky

paulijar:
I'm sorry but I'm not going to merge this PR. Defining the radius as pixels may be your personal preference but it is not an objectively better way to do this. It's not good enough that a commit "makes no harm". It should bring clear added value; otherwise it is just litter in the git history.

In your own theme you are of course free to override any css rule you wish.

PS. If anyone wonders, why this PR now contains some "[tx-robot] updated from transifex" commits, this is because I just force-pushed the master branch. There has been incorrect configuration on tx-robot and it has been pushing broken translation changes to the master branch for a few days, and this was the easiest way to get rid of those changes. I know that force-pushing on collaborative project is kind of evil but I hope this doesn't cause too much inconvenience for too many people.

PR #679, opened 2019-01-26
Russian playlist translation fix

TFBosoN:
Nobody calls playlist "список проигрывания" in Russia. "плейлист" is a modern word for that.
It's unavailable in transifex so I requested it here.


paulijar:
Okay, thanks! I hope the Transifex translation process starts functioning again for us some day, and it's possible that this kind of manual fixes get overridden at that point. But it may still take time, so I'll merge this for now.

PR #688, opened 2019-03-13
Changed image loading method for accessibility compatibility.

### LINKED ISSUE #683: Issues with the dark theme in nextcloud
All the icons and button are black and so is the background in dark mode, which makes it almost impossible to see the buttons:
image

Faldon:
Summary
------------
For images defined as src element in template files, use the Nextcloud SVG API to access the images iff the
dark theme is available and used.
For the embedded player, do the same through the javascript OC module.
To invert icons defined as background in the css stylesheet, an invisible SVG is inserted into the navigation list defining a filter to invert the colour. 
When all artists are loaded, apply the svg filter iff the dark theme is available and used.

 For Owncloud compatibility, these changes are encapsulated in a check if the accessibility app is available and 'themedark' is  configured and used as user theme.

Notes
--------
Although the Nextcloud theming app has public methods to access a themed image from an app, the dark theme in question resides in the Accessibility app (and obviously does not provide a similar api to convert app images). So, the distinction have to be made programmatically.

Quirks / Caveats
----------------------
The changes are made to be compatible with the original Owncloud appliance. But due to the nature of this bug described in Notes, a first workaround is to detect if the specific theme was chosen as user theme. If another accessible theme will be introduced, the relevant parts of the code has to be adapted again.

Fixes #683

### DIFF
--- a/js/app/controllers/alltracksviewcontroller.js
+++ b/js/app/controllers/alltracksviewcontroller.js
@@ -88,6 +88,10 @@ angular.module('Music').controller('AllTracksViewController', [
 				$scope.tracks = libraryService.getTracksInAlphaOrder();
 				$timeout(function() {
 					$rootScope.loading = false;
+					if($rootScope.darkTheme) {
+						$('#alltracks-area').find('.svg')
+							.css('filter', 'url("#backgroundInvert")');
+					}
 				});
 				setUpAlphabetNavigation();
 			}
--- a/js/app/controllers/maincontroller.js
+++ b/js/app/controllers/maincontroller.js
@@ -21,6 +21,7 @@ function ($rootScope, $scope, $timeout, $window, $document, ArtistFactory,
 
 	$rootScope.playing = false;
 	$rootScope.playingView = null;
+	$rootScope.darkTheme = (OCA.hasOwnProperty('Accessibility') && OCA.Accessibility.theme == 'themedark');
 	$scope.currentTrack = null;
 	playlistService.subscribe('trackChanged', function(e, listEntry){
 		$scope.currentTrack = listEntry.track;
@@ -72,10 +73,14 @@ function ($rootScope, $scope, $timeout, $window, $document, ArtistFactory,
 			$rootScope.loadingCollection = false;
 
 			// After data has loaded, invert icons in case the nextcloud dark theme is configured
-			if(OCA.hasOwnProperty('Accessibility') && OCA.Accessibility.theme == 'themedark') {
+			if($rootScope.darkTheme) {
 				var navigation = $('#app-navigation');
-				navigation.find('.music-nav-settings a').css('filter', 'url("#backgroundInvert")');
-				navigation.find('.play-pause').css('filter', 'url("#backgroundInvert")');
+				navigation.find('.music-nav-settings a')
+					.css('filter', 'url("#backgroundInvert")');
+				navigation.find('.play-pause')
+					.css('filter', 'url("#backgroundInvert")');
+				$(document).find('.svg')
+					.css('filter', 'url("#backgroundInvert")');
 			}
 		},
 		function(response) { // error handling
--- a/js/app/controllers/maincontroller.js
+++ b/js/app/controllers/maincontroller.js
@@ -70,6 +70,13 @@ function ($rootScope, $scope, $timeout, $window, $document, ArtistFactory,
 				$rootScope.$emit('playlistsLoaded');
 			});
 			$rootScope.loadingCollection = false;
+
+			// After data has loaded, invert icons in case the nextcloud dark theme is configured
+			if(OCA.hasOwnProperty('Accessibility') && OCA.Accessibility.theme == 'themedark') {
+				var navigation = $('#app-navigation');
+				navigation.find('.music-nav-settings a').css('filter', 'url("#backgroundInvert")');
+				navigation.find('.play-pause').css('filter', 'url("#backgroundInvert")');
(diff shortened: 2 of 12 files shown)


paulijar:
Thanks for your PR. I'll try to take a look at it during the weekend.

paulijar:
Is there maybe something missing from this PR? When I pulled the changes, I just lost all the icons when the dark theme is active:
image

For example, the play button element now looks like this
`Play`
...but no image can be found from the url `/nextcloud_15/svg/music/play-big`.

paulijar:
Thanks also for looking at the Travis issues. The builds have been failing for some time now, but I've been too busy with other things to take a closer look at the reasons behind it. It's actually quite frustrating how the Travis builds keep breaking even if we do nothing; we need to keep running just to stay still.

Anyway, I now committed Travis fixes to the master branch based on your example, as I wanted to fix it before this PR is ready to be merged. You should now remove the Travis changes from this PR as they are now in conflict with the master, and no longer needed.

Faldon:
Hmmm.... My first guess was a caching issue. In my development server, all looks fine:
oc_music_dark_theme

But the image source confuses me. 
`Pause`
The nextcloud_15 prefix in the src path I can't explain. Are all link to apps prefixed with nextcloud_15 for you?

Faldon:
@paulijar I rebased my branch to remove the travis fix. Hope, I did it the right way. :open_mouth:

paulijar:
> The nextcloud_15 prefix in the src path I can't explain. Are all link to apps prefixed with nextcloud_15 for you?

My cloud instance is reached via address `https://<domain>/nextcloud_15`. When the dark theme is not active, the same element looks like this:
`Play`.

It shouldn't be a caching issue, either. I tested two browsers (Firefox and Chrome) and made many forced reloads with cache clearing.

> I rebased my branch to remove the travis fix. Hope, I did it the right way. 😮

The Travis fix was removed correctly, but your branch is still sitting on top of an older commit of the master branch which is why github is now complaining that "Some checks were not successful". You should rebase your branch on the latest HEAD commit and make another force push.

Faldon:
Ok, therefore the prefix. But this is somewhat weird. 
I use two different methods to fetch the icon, wether the dark theme is enabled or not. So, the links are correct in both themes.
Could you try to enter the link to an svg (e.g. `https://<domain>/nextcloud_15/svg/music/pause-big`)  in the browser and see, if it fetches the icon?
Rebased onto master, now the checks should succeed.

paulijar:
> Could you try to enter the link to an svg (e.g. `https://<domain>/nextcloud_15/svg/music/pause-big`) in the browser and see, if it fetches the icon?

Yes, I tried this before and now again, and this returns just "404 Not found". I'm testing on freshly installed Nextcloud 15.0.5 with no other applications except Music installed if that matters.

Do you know some other applications which use similar strategy to load their icons? Maybe I could test if those work on my installation.

Faldon:
No, I'm not aware of. But it just came to mind, maybe it fails when pretty urls are configured.
I currently set a development environment up as I find it in general in production.
If this method won't work well, maybe switching to inverting the icons (like in the navigation pane) in javascript would be a better approach.

Faldon:
Ok, I think I found the error. Do you have your urls with the index.php visible in the path? As far as I can see, when the base path contains respectively ends with index.php, than the link is not generated correct.
But, as I already have to use dynamic javascript to inverse the icons in the navigabtion bar, shall I adopt this method to the other icons as well and thus revert the other modifications? 
It has the advantage of cleaner code, because there will be only one location, where a "theme detection" has to be made.

paulijar:
> Do you have your urls with the index.php visible in the path? As far as I can see, when the base path contains respectively ends with index.php, than the link is not generated correct.

Yes, that's right, I have the index.php visibile on the addresses.

> But, as I already have to use dynamic javascript to inverse the icons in the navigabtion bar, shall I adopt this method to the other icons as well and thus revert the other modifications?

Yes, this sounds like a good plan. I would have anyway requested you to make the code more [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) as currently the following piece of logic is repeated a zillion times within the templates:
`<?php $darkThemeEnabled ? p(link_to('svg', 'music/play-big')) : p(OCP\Template::image_path('music', 'play-big.svg')); ?>`
I was thinking about creating a single PHP function which would have been called like
`<?php musicImagePath('play-big')?>`
...but doing all the inverting within js will take care of the problem, too.

Faldon:
@paulijar Ok, so now the icon coloring is made in the js app. I did not find the perfect place to have it a one liner, but had to split it up in three parts. However, it's much cleaner code now, imho.

paulijar:
Okay thanks, now it works for me, too. I think this is now good enough for merging although I probably try to find even cleaner solution if I have spare time at some point.

Also, this doesn't solve all the issues with the dark theme. E.g. the info note shown for empty playlist or during scanning still has white background, and so does the embedded files player. But this is a step forward and we don't need to fix everything in one PR.

Faldon:
@paulijar Sounds great. I think, you currently have a better understanding, when certain parts of the javascript code is loaded and executed.

paulijar:
FYI: I think I found a rather nice solution for the dark theme support. It's now implemented in the commit 4c50beee0. Also, I made the support more comprehensive with commits 0546471a8e and 48456580a6.

Faldon:
@paulijar Oh, yes. This is indeed much cleaner with the code for the styling where it belongs to.

Issue #693, opened 2019-04-06
exception when render JSONResponse fileinfo details pane


greku:
I have an issue with some of my audio files when opening the details pane.
The server returns http code 500 when requesting the file details.
The error is caused by an exception in JSONResponse encode: https://github.com/nextcloud/server/blob/2e36069e24406455ad3f3998aa25e2a949d1402a/lib/public/AppFramework/Http/JSONResponse.php#L71

This seems to be due to strange characters present in the fileinfo/encoder string. (see attached picture)
![issue_fileinfo_mp3](https://user-images.githubusercontent.com/259454/55674893-1373de00-58bb-11e9-9696-1b008bd70ebd.png)

I have music app version 0.9.3, nextcloud 13.

paulijar:
Thanks for the report. I believe we should add sanitazing of the details data to about here: https://github.com/owncloud/music/blob/master/utility/detailshelper.php#L42.

This kind of thing already happens for the fields which are added to the database, but this path is is not executed when getting data for the details data: https://github.com/owncloud/music/blob/master/utility/extractorgetid3.php#L84.

greku:
Thanks for the guidance. This seems to solve the problem on my setup. Please review the change too.

paulijar:
The fix is now released in version 0.9.5

PR #694, opened 2019-04-10
sanitize strings of audio file details before encoding #693

greku:


utility/detailshelper.php
```diff
@@ -36,7 +36,8 @@ public function getDetails($fileId, $userFolder) {
 		$fileNodes = $userFolder->getById($fileId);
 		if (\count($fileNodes) > 0) {
 			$data = $this->extractor->extract($fileNodes[0]);
-
+			array_walk($data['audio'], [$this, 'sanitize_string']);
+			array_walk($data['comments'], [$this, 'sanitize_string']);
```
paulijar: This call probably does nothing because $data['comments'] does not contain any strings, it contains arrays. That's why there's the self::flattenComments call below.

utility/detailshelper.php
```diff
@@ -83,6 +84,16 @@ private static function encodePictureTag($pic) {
 		}
 	}
 
+	/**
+	 * Remove potentially invalid characters from the string.
+	 * @param $item
+	 */
+	private function sanitize_string(&$item){
```
paulijar: This function could be made static

greku: I have added static keyword.

utility/detailshelper.php
greku: I have updated the order of the operations to solve this issue.

paulijar: Thanks. I have two more minor stylistic requests.
1. Instead of `array_walk($something, [$this, 'sanitize_string'])`, you could achieve the same with `array_walk($something, 'self::sanitize_string')`. I feel this would be slightly cleaner.
2. All the PHP codes of the Music app use camelCase instead of snake_case. So the function name should actually be `sanitizeString`.

greku: You're right. Updated for consistent style.

greku: The inspection doesn't understand that the method is called with this code:
`array_walk($something, 'self::sanitize_string')`
I would suggest to revert to the original one then? `array_walk($something, [$this, 'sanitize_string'])`

paulijar: I'll let you decide, I'm content either way. On one hand, it's good to avoid creating Scrutinizer warnings. On the other hand, I dislike a bit the idea that we let false positives from an analysis tool to control our coding.

greku: scrutinizer doesn't really detect it correctly, even with the full class name. so I suggest to keep it as it is now.

Issue #696, opened 2019-04-13
alphabet navigation artists/albums starting by number


greku:
I have artist that start with a number like "3...". in that case, it is sorted before 'A', but clicking on the first element of the alphabet link "A" doesn't go to the top of course,  and therefore, not up to this artist "3...". My suggestion is to add '#' symbol before A, that goes to the top of the list.
See: https://github.com/greku/music/tree/alpha_nav_num

paulijar:
Solved by #697

PR #697, opened 2019-04-13
alphabetical navigation, add # before A for items starting with number

greku:
implementation according to suggestion in issue #696


paulijar:
Makes sense and seems to work well. One minor change is needed, though: In both controllers, the comparison `prevLetter==''` should be `prevLetter===''`. It makes no functional difference in this case, but jshint is complaining about the former, at least on my system. This prevents making the app.js.

greku:
I fixed it. I don't know why, but jshint does not report anything on my pc and the app.js just builds.

paulijar:
Okay, thanks. Maybe this depends on the version of jshint used (?). Anyway, now this is good for merging.

PR #699, opened 2019-04-24
Alphabet navigation, grouped letters links for small height

greku:
Letters are grouped by two instead of showing one letter out of two.
Grouped letters links to the first available item starting with one of the two letters.

### DIFF
--- a/css/public/app.css
+++ b/css/public/app.css
@@ -177,10 +177,6 @@
 	display: none;
 }
 
-.alphabet-navigation a.filler.stripped
-{
-	display: none;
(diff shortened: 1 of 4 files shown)


paulijar:
Works well, thanks again for your contribution!

paulijar:
FYI: I added another grouped layout for extra short screens with commit 07f1d584e3. Having 15 links on small landscape smart phone made the layout too crowded, so I reduced this to 9 links when the height of the alphabet navigation bar is less than 200 pixels.

PR #744, opened 2020-03-03
Support year for .m4a

### LINKED ISSUE #743: Covers don’t show for albums with m4a files
Since I updated to v0.13.0 covers of files with the `.m4a` extension (AAC encoding) don’t show up in Album View, while they used to in previous versions. It doesn’t seem to matter if there’s a `cover.jpg` in the folder. If it’s an m4a, it doesn’t show up.
Covers of mp3 and flac files still work. 
I noticed there is a difference in the HTML source code between `.m4a` and `mp3`. M4a looks like this

```
/apps/music/api/album/13096/cover
```
Which returns an empty JSON. And mp3 looks like this:

```
/apps/music/api/cover/f1c9f1ea52953f6492e01f62fb1a0d57
```
Which returns an image, as expected.

ChrisJAllan:
Adds year support for .m4a files.  
#743 has been fixed upstream, so second commit removed.

### DIFF
--- a/utility/scanner.php
+++ b/utility/scanner.php
@@ -230,7 +230,7 @@ class Scanner extends PublicEmitter {
 		$meta['discNumber'] = self::normalizeOrdinal($meta['discNumber']);
 
 		// year
-		$meta['year'] = ExtractorGetID3::getFirstOfTags($fileInfo, ['year', 'date']);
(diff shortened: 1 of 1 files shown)


paulijar:
Thanks for the PR. I'm really busy for the next two weeks, so I didn't have time to test this. But as the change is so small and simple, I believe it's safe to merge this.

PR #806, opened 2020-11-23
I hate dead links

bmschwa:
I didn't see an issue template in this repo, so I just referenced back to the core project.


paulijar:
Thanks! I probably should write my own issue template at some point, though. I feel that the ownCloud core template is rather cumbersome to fill and it contains lot of fields rarely relevant for the Music app bugs. But it's not nice to have dead links, either.

PR #888, opened 2021-08-31
set user agent when fetching podcasts

icewind1991:
Requests without valid user agents are blocked by some cdn services.

Currently trying to add "https://feeds.acast.com/public/shows/6046295bba4c4c35b8b757cb" (and I assume any other podcast hosted on the same platform) will result in an "Invalid url" error.

(Thanks for adding the podcasting feature, I was just looking for a good way to listen to them when the update was released :heart: )


paulijar:
Thanks for the PR, much appreciated!

paulijar:
This is now released in Music v1.3.2

PR #890, opened 2021-09-01
Update German translation

simonspa:
Here a complete translation of the app into German, using the formal "Sie" form where direct address is necessary. Since the translation was terribly outdated, I have added the same translation for all German variants, tuning may be done later. :)

### DIFF
--- a/l10n/de/music.po
+++ b/l10n/de/music.po
@@ -9,7 +9,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: ownCloud\n"
 "POT-Creation-Date: \n"
-"PO-Revision-Date: 2021-09-01 21:04+0200\n"
+"PO-Revision-Date: 2021-09-01 21:53+0200\n"
 "Last-Translator: Simon Spannagel <simonspa@kth.se>\n"
 "Language-Team: German (Germany) (http://www.transifex.com/owncloud-org/"
 "owncloud/language/de_DE/)\n"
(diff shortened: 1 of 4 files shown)


simonspa:
Ooops, thanks for checking so diligently, this somehow escaped my attention in the Poedit interface. Here you go!

paulijar:
Looked good to me now. Thanks again for the good work!

paulijar:
This is now released in Music v1.3.2

raimund-schluessler:
The German informal translation ("du") does not work. I suspect this is because the language is given as de_DE in
https://github.com/owncloud/music/blob/e16209ddb935cd7de4186e9f5fd8b3ed15ea6bc9/l10n/de/music.po#L16

The same might be true for de_AT and de_CH. @simonspa can you confirm this?

simonspa:
I can confirm that the German informal language isn't properly picked up. @raimund-schluessler where can I find an example of how this *should* be set in order for the selection to work?

raimund-schluessler:
> 
> 
> I can confirm that the German informal language isn't properly picked up. @raimund-schluessler where can I find an example of how this _should_ be set in order for the selection to work?

I don't know, to be honest. Maybe just `de` (and `de_CH`, `de_AT`)?

paulijar:
> I don't know, to be honest. Maybe just de (and de_CH, de_AT)?

Yes, this is correct. I just made a fix like this. See the commit above.

As a side note, the Switzerland variant de_CH is probably never used; I couldn't find such language option Neither from Nextcloud nor ownCloud. The Austrian variant de_AT *can be* selected on ownCloud but not on Nextcloud.

paulijar:
Music v1.4.0 is now out and it should show these translations on all German variants.

simonspa:
I can confirm, works like a charm now - thanks @paulijar for fixing!

Issue #898, opened 2021-09-26
zh_cn is not fully translated.


RuofengX:
![图片](https://user-images.githubusercontent.com/40327863/134794410-5f6ffa07-6aaf-47bf-9bbb-b636a890834c.png)
As we can see in this image, this project is not fully translated. If I want to make some translation, what should I do? I notice there's a i10n folder, could someone gives me a tip?

paulijar:
Yes, the same is true for the most of the languages. See https://github.com/owncloud/music/issues/671#issuecomment-782746463 for instructions on how to contribute translations.

paulijar:
This has been resolved by https://github.com/owncloud/music/pull/899.

PR #899, opened 2021-09-26
Updated zh_CN/music.mo

RuofengX:
Update zh_CN/music.mo from templete/music.mo, and add some translation. Try to solve issue #898


RuofengX:
Try to solve issue #898

RuofengX:
Some errors appear when trying to generate` js/app/l10n/translations.js`, I never used npm before. So the pr didn't countain translation.js file.
Here is the full log, if someone needed:  
```
ubuntu@cloud:~/music/music/build$ sudo make l10n-compile
npm install --deps
npm WARN optional SKIPPING OPTIONAL DEPENDENCY: fsevents@2.3.1 (node_modules/fsevents):
npm WARN notsup SKIPPING OPTIONAL DEPENDENCY: Unsupported platform for fsevents@2.3.1: wanted {"os":"darwin","arch":"any"} (current: {"os":"linux","arch":"x64"})
npm WARN optional SKIPPING OPTIONAL DEPENDENCY: fsevents@1.2.13 (node_modules/watchpack-chokidar2/node_modules/fsevents):
npm WARN notsup SKIPPING OPTIONAL DEPENDENCY: Unsupported platform for fsevents@1.2.13: wanted {"os":"darwin","arch":"any"} (current: {"os":"linux","arch":"x64"})

audited 908 packages in 5.608s

70 packages are looking for funding
  run `npm fund` for details

found 1 moderate severity vulnerability
  run `npm audit fix` to fix them, or `npm audit` for details
npm run l10n-compile

> music@1.3.3 l10n-compile /home/ubuntu/music/music/build
> angular-gettext-cli --compile --files ../l10n/**/music.po --dest ../js/app/l10n/translations.js --module Music

Finished compiling ../l10n/ach/music.po, read 1 files.
All files processed, stored in a file ../js/app/l10n/translations.js
cd ../l10n; ./l10n.pl music write
Can't locate Locale/PO.pm in @INC (you may need to install the Locale::PO module) (@INC contains: /etc/perl /usr/local/lib/x86_64-linux-gnu/perl/5.30.0 /usr/local/share/perl/5.30.0 /usr/lib/x86_64-linux-gnu/perl5/5.30 /usr/share/perl5 /usr/lib/x86_64-linux-gnu/perl/5.30 /usr/share/perl/5.30 /usr/local/lib/site_perl /usr/lib/x86_64-linux-gnu/perl-base) at ./l10n.pl line 3.
BEGIN failed--compilation aborted at ./l10n.pl line 3.
make: *** [Makefile:44: l10n-compile] Error 2
```

paulijar:
Thanks! The error seems to happen after the translations.js is created. The script tries to create also localization files usable on the back-end, but this now fails because some Perl module is missing. But don't worry, I can do the technical steps.

I spotted one thing in the translations which would still need fixing. These strings:
image
have both singular form and plural form, although only the singular is shown in the list view of Poedit. But in the bottom of the screen, you should see both singular and plural forms and both translations should be supplied:
image

RuofengX:
Sorry, I missed some string somehow. It's been added now if nothing went wrong.

paulijar:
Sorry, this still doesn't look quite right. Your second commit added the music.mo binary file but we don't need that for anything. We are only interested about the music.po file, and that didn't change at all in your second and third commits.

RuofengX:
Oh I confused with mo and po file. I will fix it later

RuofengX:
OK, music.po file is up-to-date and music.mo is deleted. Maybe a rebase job is needed, but I don't know how to

paulijar:
Thanks, it's better but still doesn't address my original comment. If you look at the "Files changed" tab in github, it can be easily seen in the raw music.po file. For example, there is this block:
```
#: ../js/app/controllers/maincontroller.js:54
msgid "1 album"
msgid_plural "{{ count }} albums"
msgstr[0] "专辑"
```
There's supposed to be two translations, one for the case of only one album in the user's library and another one for the case when there are more than one. The latter translation should contain the placeholder `{{ count }}` which will get replaced with the numeral showing the actual count. In practice, 99+% of the cases will be such that there are more than 1 albums. This translation will be used as a tooltip, when the user is hoovering over the "Albums" links on the navigation pane.

Using Poedit, you should fill the translations of "1 album" on the tab "Singular" and the translation of "{{ count }} albums" on the tab "Plural".

RuofengX:
Actually mandarin doesn' t have a plural form. The word 专辑 means both album and albums. Also in mandarin, it is better add a unit before a object, let's say, 1张专辑, the word 张 means one or some pieces of album. And because of the dependence issue,  I can't tell whether I need to add those unit words,  so I would make more change after a build.
 I also have a problem about the placeholder. I meet 2 kinds  of placeholder, one is {{ count }},  Another is {{count}},  Which one is correct?

RuofengX:
> Using Poedit, you should fill the translations of "1 album" on the tab "Singular" and the translation of "{{ count }} albums" on the tab "Plural".

I didn't notice the Singular and Plural tab,  I just translate both 1 album and {{ count }} albums into 专辑, because they are the same thing in mandarin.  Maybe I miss the place holder, I'll check it.

RuofengX:
Yes, the placeholder somehow missing, I will check it later.

> Using Poedit, you should fill the translations of "1 album" on the tab "Singular" and the translation of "{{ count }} albums" on the tab "Plural".

RuofengX:
I think I find the problem. From [Baidu ZhiDao ](https://zhidao.baidu.com/question/1047426001800427259.html)
图片
The poedit program thinks that mandarin do not need a Plural form, because Single and Plural are same, so I can't see a Plural tab, instead, it only shows a tab called 一切, means any.  
Also, `{{ count }} albums` is not the Plural form of `1 album` that is the problem.  
My advise is use two different string to store these different concept, or, I add the "Plural" form directly in plain file.

paulijar:
Okay, if that is the case, then would it make sense to translate it as "{{ count }} 专辑" or something similar? And the  {{ count }} there could then be number 1 or any greater number? I still need to check, if the application works technically correctly when only one translation is given like this.

RuofengX:
In old version, those two string are two different string, I think that is the best solution.  
I want to use text editor to add the string directly. So the `msgstr[0] "1张专辑"`is for singular form and `msgstr[1] "{{ count }}张专辑"`  is for plural form? If that' s correct I could addit directly not using poedit.  This could solve the problem temporarily with no change to upstream.

RuofengX:
Also I think the string `1 album` is only meaningful in English and other language with plural form. It is same to {{ count }} album in Chinese, because album is same to albums in chinese

RuofengX:
I updated the po file using things like `msgstr[1] "{{ count }}张专辑"`.

paulijar:
I did now some testing, and the correct way seems to be to provide just one translation like `"{{ count }}专辑"`. If we provide both singular and plural forms like in your latest commit, then the singular `"1专辑"` is always pulled by the translation library, regardless of the actual count.

In case you are interested, this is how we ask for the translated string in the code:
https://github.com/owncloud/music/blob/599a1bef7a5a0e6b9878f6f7b281fb1c6ff23279/js/app/controllers/maincontroller.js#L54
Here, the gettextCatalog library automatically decides, which of the singular/plural translations to use, depending on the first argument `albumCount`. I guess that gettextCatalog also internally knows that there is supposed to be only one form for Chinese which is then used regardless of the count. I have read that gettextCatalog supports also languages using three different forms (1 item, 2 items, or 3+ items); e.g. Slovenian is such a language.

While testing, I also noticed one more thing to fix. In this translation, the parameter names and count do not match:
```
#: ../templates/partials/views/settingsview.php:156
msgid ""
"Array of allowed non-HLS radio streaming hosts. Default is ['http://*:*', "
"'https://*:*'], allowing streaming from any remote URL. The given URLs will "
"be added to the Content-Security-Policy header <a href=\"{{url}}\" target="
"\"_blank\">media-src</a>."
msgstr ""
"允许的非HLS电台流媒体主机列表。 默认值为['http://*:*', 'https://*:*']，允许来"
"自任何远程URL的流。 给定的URL将添加到标头Content-Security-Policy中<a href="
"\"{{connectSrcUrl}}\" target=\"_blank\">connect-src</a>。 此外，如果您指定任"
"何允许的来源，则来源<samp>data:</samp>和<samp>blob:</samp>也将添加到 CSP<a "
"href=\"{{mediaSrcUrl}}\" target=\"_blank\">media-src</a>。"
```
Also, that translation refers to the `data:` and `blob:` URLs which are not there in the original. They are part of the other quite similar string, but not this one.

Other than this, everything seemed to work technically correctly, and I saw no missing translations.

paulijar:
As a mildly interesting technical detail, this line in music.po controls, which kind of singular/plural/etc. translations can be entered with Poedit:
https://github.com/owncloud/music/blob/599a1bef7a5a0e6b9878f6f7b281fb1c6ff23279/l10n/zh_CN/music.po#L18

In the Finnish translation file, the corresponding line is like this:
https://github.com/owncloud/music/blob/599a1bef7a5a0e6b9878f6f7b281fb1c6ff23279/l10n/fi_FI/music.po#L17

And in the Slovenian file, it's like this:
https://github.com/owncloud/music/blob/599a1bef7a5a0e6b9878f6f7b281fb1c6ff23279/l10n/sl/music.po#L14

But none of these definitions are migrated to translations.js, so the gettext library must have corresponding/compatible definitions built-in and hard-coded per language.

RuofengX:
Thanks for those careful test. I still need to confirm that:  
> I did now some testing, and the correct way seems to be to provide just one translation like `"{{ count }}专辑"`. If we provide both singular and plural forms like in your latest commit, then the singular `"1专辑"` is always pulled by the translation library, regardless of the actual count.

So only plural form `msgstr [0] {{ count }}` is needed, because zh_CN.po file only *provide* the one form everytime. is that right?

---



> Also, that translation refers to the `data:` and `blob:` URLs which are not there in the original. They are part of the other quite similar string, but not this one.  
I made a mistake,  thanks for notice!

paulijar:
Thanks, looks correct now. I'll test this one more time in the evening, and then merge if I find no more issues.

paulijar:
Everything worked great now, so I squashed and merged the commits. Thanks again for your contribution, it's much appreciated! The translations will be included in the next Music release v1.4.0 which I hope to be able to publish within a couple of weeks.

paulijar:
The Music v1.4.0 with the Chinese translations is now released.

PR #900, opened 2021-09-26
Update README.md

RuofengX:
Travis-ci.org says that 'Since June 15th, 2021, the building on travis-ci.org is ceased. Please use travis-ci.com from now on.'

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # README
 
-[![Build Status](https://secure.travis-ci.org/owncloud/music.png)](http://travis-ci.org/owncloud/music)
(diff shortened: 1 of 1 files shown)


RuofengX:
Changing only `org` to `com` just didn't work.

paulijar:
Yeah, Travis combined their OSS oriented travis-ci.org to travis-ci.com targeted for paying customers. When they did this, they promised to keep offering free builds for the open source community. But that was bullshit, there is now a "build credit" system to run the builds and there is a system to give free credits for the open source projects. But they haven't actually given a single free credit for months now. So what we should really do, would be to ditch Travis and set up some CI tests using github actions.

PR #912, opened 2021-11-04
Enable for Nextcloud 23

PVince81:


PVince81:
I did a quick smoke test with NC 23.0.0 beta3:
- music playback in web UI still works
- music playback with Ultrasonic app still works

paulijar:
Great, thanks!

paulijar:
This is now released in Music v1.5.0.

PR #957, opened 2022-03-25
Allow app for Nextcloud 24

PVince81:
I haven't tested this yet and might do this in the next days.


PVince81:
Navigating in the web UI and streaming from there works fine.

Streaming from the Ultrasonic Android app seems to work fine as well.

paulijar:
Thanks for the PR and testing!

paulijar:

PR #970, opened 2022-04-29
Fixed XML validation error with C# like clients like MusicBee

rstefko:


paulijar:
Great, thanks. I rebased and merged this manually, so I'm just going to close this PR.

Btw, does the connection between MusicBee and ownCloud Music work properly for you after this change. Because for me, this change does help me to get a bit further than before, but then the MusicBee [Subsonic plugin](https://getmusicbee.com/addons/plugins/41/subsonic-client/) gets stuck in the phase "Running GetMusicFolders". I wonder if there is also some other problem in our Subsonic API implementation or if there is just some property in my library specifically which reveals an issue on MusicBee.

rstefko:
The MusicBee plugin relies on path attribute which is not supported by ownCloud music. But as I was looking into the code, it feels like it is better to make changes in the MusicBee plugin: https://github.com/midwan/MB_SubSonic/pull/66

paulijar:

Issue #967, opened 2022-04-22
PodcastEpisodeBusinessLayer fails to parse MM:SS duration format


ksmolder:
When importing a particular podcast (), I noticed that Music didn't extract the duration of the episodes. Looking at the `PodcastEpisodeBusinessLayer` code, Music currently supports two duration formats:

- Total number of seconds
- HH:MM:SS

The episodes that failed to have their duration extracted, had the hours part omitted from the HH:MM:SS representation, so only the minutes and seconds were included.

The following regular expression would be more flexible and support also the MM:SS format:
 `^(?:(?:(?<hours>\d+):)?(?<minutes>[0-5]?\d):)?(?<seconds>[0-5]\d)(?:\.\d*)?$`

The downside is that it looks frightening ;-), so that's why I included the named references.

Recognized duration formats:

- 12:45:56 → 12 hours + 45 minutes + 56 seconds
- 12:45:78 → Not recognized as the number of seconds is higher than 59
- 12:45 → 12 minutes + 45 seconds
- 12 → 12 seconds
- 12.9 → 12 seconds
- 12:45.67 → 12 minutes + 45 seconds

So basically, subseconds are still discarded.

The inclusion of named groupings might make the actual calculation more readable (to counteract the more difficult regex). The `parseDuration` method would read:

    private static function parseDuration(string $data) :?int {
        $matches = null;
    
        if (\ctype_digit($data)) {
            return (int)$data; // plain seconds
        } elseif (\preg_match('/^(?:(?:(?<hours>\d+):)?(?<minutes>[0-5]?\d):)?(?<seconds>[0-5]\d)(?:\.\d*)?$/', $data, $matches) === 1) {
            return (int)$matches['hours'] * 3600 + (int)$matches['minutes'] * 60 + (int)$matches['seconds']; // HH:MM:SS
        } else {
            return null; // no value or unsupported format
        }
    }

An alternative approach could be to add an additional `elseif` block to the check, especially checking for the MM:SS format. The downside of this approach is that if a podcasts includes the duration as SS.mmm, none of the checks would succeed.

    private static function parseDuration(string $data) :?int {
        $matches = null;
    
        if (\ctype_digit($data)) {
            return (int)$data; // plain seconds
        } elseif (\preg_match('/^(\d\d):(\d\d).*/', $data, $matches) === 1) {
            return (int)$matches[2] * 60 + (int)$matches[3]; // MM:SS
[...]

PR #971, opened 2022-05-01
Updates the regexp used to parse a podcast's episode's duration

ksmolder:
The updated regular expressions supports the following duration formats:

* HH:MM:SS
* HH:MM:SS.sss
* MM:SS
* MM:SS.sss
* SS
* SS.sss

Accepted values:

* HH: 0 -> infinite. As we calculate a duration, the number of hours is not limited
* MM: 0 -> 59 (Both 00, 01,... and 0, 1,... are recognized)
* SS: 0 -> 59 (Both 00, 01,... and 0, 1,... are recognized)

If the subsecond part is present, it is neglected when caculating the episode's duration.

Resolves owncloud/music#967


paulijar:

PR #987, opened 2022-06-15
Add webradio stream title with icecast/shoutcast

medismail:
Add support for webradio stream title

### DIFF
--- a/appinfo/app.php
+++ b/appinfo/app.php
@@ -76,6 +76,7 @@ function adjustCsp(IAppContainer $container) {
 
 		foreach ($radioSources as $source) {
 			$policy->addAllowedMediaDomain($source);
+			$policy->addAllowedConnectDomain($source);
 		}
 
 		foreach ($radioHlsSources as $source) {
--- a/js/shared/icecastMetadata.js
+++ b/js/shared/icecastMetadata.js
@@ -1,0 +1,176 @@
+/**
+ * ownCloud - Music app
+ *
+ * This file is licensed under the Affero General Public License version 3 or
+ * later. See the COPYING file.
+ *
+ * @author Moahmed-Ismail MEJRI <mmismailm@hotmail.com>
+ * @copyright Moahmed-Ismail MEJRI 2022
+ */
+
+import IcecastReadableStream from './icecast-metadata-js/IcecastReadableStream.js';
+
+OCA.Music = OCA.Music || {};
+
+OCA.Music.ReadStreamMetaData = function(endpoint, onMetaDataCb, method) {
+    var StreamSource = {
+        SHOUTCAST_V1: 'SHOUTCAST_V1',
+        SHOUTCAST_V2: 'SHOUTCAST_V2',
+        STREAM: 'STREAM',
+        ICECAST: 'ICECAST'
+    };
+
+    function getShoutcastV1Station(url, callback) {
+        url = url + '/7.html';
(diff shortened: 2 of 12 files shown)


paulijar:
Thanks! I'll check this out as soon as I have time. One quick comment though: icecast-metadata-js is a 3rd party library and it should be available as an NPM package. So there should be no need to copy it to our version control. It should be sufficient to run `npm i icecast-metadata-js` in the `build` folder and then require the new library in https://github.com/owncloud/music/blob/master/js/index.app.js.

On the other hand, if a 3rd party library couldn't be obtained as a node module, then it should reside under `js/vendor` instead of `js/shared`.

medismail:
Thank you for review, icecast-metadata-js can be added from npm this is the repo: https://github.com/eshaz/icecast-metadata-js/tree/master/src/icecast-metadata-js#readme
I didn't find how to add it properly so I copy the files.
I think also I have a issue to update media session each time the title change, because the $scope.$watch for CurrentTrack not start when I change  CurrentTrack.currentTitle.

appinfo/app.php
```diff
@@ -76,6 +76,7 @@ function adjustCsp(IAppContainer $container) {
 
 		foreach ($radioSources as $source) {
 			$policy->addAllowedMediaDomain($source);
+			$policy->addAllowedConnectDomain($source);
```
paulijar, line 79: This is somewhat problematic because by default, this now makes a rather large hole to the CSP. The default allowed radio sources is *:* so after this, any javascript running within ownCloud/Nextcloud is allowed to connect which ever URL it likes. This basically undermines the whole point of the Content Security Policy. Also, this is the reason, why I require each cloud administrator to whitelist the HLS radio sources they like to allow: Playing the HLS streams requires this connect-src CSP rule, and I don't want to automatically weaken the security of each and every cloud instance running this app.

Maybe the solution might be to relay all this metadata loading via our own server? So the Music app PHP backend would have a REST API to get the radio metadata and its implementation would make the calls to the icecast/shoutcast servers. I believe this would help also with the cases where the icecast/shoutcast server doesn't define the needed CORS headers.

js/shared/icecastMetadata.js
```diff
@@ -0,0 +1,176 @@
...
+                options
+            );
+
+            await icecast.startReading();
+        });
+    }
+
+    var methodHandler = undefined;
+
+    switch (method) {
```
paulijar, line 157: Apparently the code is now always using the "STREAM" method? Is there any reason to have these others here? In which case would they be useful?

appinfo/app.php
medismail: I agree that with php it will be easier and more secure, I was thinking that using the server to load information is more bad, if you agree I will try new method with php in the server.

paulijar: Please do. That shouldn't generate too much server load as the metadata is loaded with rather limited rate.

PS. I'll be busy with other things for the next week, so I probably can't give any more feedback during the next ~7 days. But I will review any updates as soon possible.

js/shared/icecastMetadata.js
medismail: Yes, I was preparing add support to for shoutcast stream title using url parsing, most of radios today use icecast metadata so is not important.

paulijar:
superseded by https://github.com/owncloud/music/pull/992

PR #992, opened 2022-06-26
Show radio stream title using php and rest api

medismail:


paulijar:
Thanks, much better now. There are still a few things which I would like to see improved, but I'll go into the details later. Please also check the things reported in the Scrutinizer inspection.

paulijar:
> This is not really an issue

You could work around it by returning the metadata as JSON. Something like `return new JSONResponse(['title' => $response]);`. And that would also be more in line with how our REST API works in general.

lib/Utility/RadioMetadata.php
```diff
@@ -0,0 +1,128 @@
...
+								fread($fp, 1);
+							}
+
+							$meta_length = ord(fread($fp, 1)) * 16;
+							if ($meta_length) {
+								$metadatas = explode(';', fread($fp, $meta_length));
+								$metadata = self::findStr($metadatas, "StreamTitle");
+								if ($metadata) {
+									$streamTitle = trim(explode('=', $metadata)[1], "'");
+									if (strlen($streamTitle) > 256) {
```
paulijar, line 106: Is there some particular reason to discard any title longer than 256 characters? Or would it make more sense to truncate those? The function `OCA\Music\Utility\Util::truncate` may be used for that.

lib/Utility/RadioMetadata.php
```diff
@@ -0,0 +1,128 @@
...
+				$out .= "User-Agent: OCMusic/1.52\r\n";
+				$out .= "Icy-MetaData: 1\r\n";
+				$out .= "Connection: Close\r\n\r\n";
+				fwrite($fp, $out);
+				stream_set_timeout($fp, $timeout);
+
+				$header = fread($fp, 1024);
+				$headers = explode("\n", $header);
+
+				if (strstr($headers[0], "200 OK") !== false) {
```
paulijar, line 86: Really minor, but the PHP manual recommends using `strpos` instead of `strstr` if you don't actually need the substring returned by `strstr`.

paulijar (changes requested):

paulijar (approved):

lib/Controller/RadioApiController.php
```diff
@@ -201,4 +203,26 @@ public function resetAll() {
...
+	*
+	* @NoAdminRequired
+	* @NoCSRFRequired
+	*/
+
+	public function getRadioStreamData(int $id) {
+		try {
+			$response = "";
+			$station = $this->businessLayer->find($id, $this->userId);
+			$stapi = $station->toAPI();
```
paulijar, line 218: One more minor thing. Converting the station to the "API format" here is quite pointless. We are interested only about the stream URL and that may be obtained with `$station->getStreamUrl()`. It should be safe to assume this to be non-null since one cannot create a station without passing a stream URL.

paulijar:
Do you consider this now to be ready for merging, or are you still planning to make any modifications?

medismail:
Yes it is ready to merge.

paulijar:
This PR, along with a lot of further development built on top of it, has now been released in Music v1.6.0.

PR #1002, opened 2022-08-19
removed white background color from controls as this makes it almost …

Erdbeergeist:
removed the white background color from the control as this makes it impossible to see the controls and the artist/album name see below:
image

### DIFF
--- a/css/music-controls.css
+++ b/css/music-controls.css
@@ -16,7 +16,6 @@
 	top: 45px;
 	position: fixed;
 	overflow: visible;
-	background-color: rgba(255, 255, 255, .93);
 	z-index: 1;
 }
 


paulijar:
We can't make it like that because then the app would look like this on NC25 with the default theme:
image
That is, the controls pane background will be completely transparent.

I guess you are using some 3rd party theme, which one? Maybe the fix should go to the theme. Or then we need to figure out a more intelligent way to do this.

Does the color of the control pane work nicely on that theme if you declare `background-color: var(--color-main-background-translucent);` for it? But even if this works, this is not yet a complete solution because this will not work on any version of ownCloud.

Erdbeergeist:
Ok I see your point, the theme is breeze dark [https://github.com/mwalbeck/nextcloud-breeze-dark](url). Setting `background-color: var(--color-main-background-translucent);`
seems to work well. I have found that there is actually a custom styling option in the theme so using that to override the default background color works well too. Maybe this is not an issue, but then in other themes this might not work. Sorry for my ignorance but can you elaborate why your suggested wont work on all versions of ownCloud. Thank you !

paulijar:
> Sorry for my ignorance but can you elaborate why your suggested fix wont work on all versions of ownCloud.

That's because ownCloud doesn't declare those color properties, that's a Nextcloud-specific thing. However, I just learned that the css `var()` syntax allows giving a default value for the cases where the named property is not found. So we could use `background-color: var(--color-main-background-translucent, rgba(255, 255, 255, .95));`. Then, the theming on Nextcloud should work as intended without breaking the ownCloud default theme.

paulijar:
I now committed a change like described above. In addition to the controls pane, I applied it on the background of the "scanning"/"no content available" banner as well as on the "embedded player" used within the Files app. The Breeze Dark theme already seemed to have an overriding rule for the embedded player background which is why no problem was seen there. But for the "scanning" banner, it uses the Music app's color definition.

Erdbeergeist:
perfect, thank you kindly :)

paulijar:
The fix has been now released in Music v1.7.0.

Issue #1004, opened 2022-08-22
SeekBar is difficult to use on certain display resolutions


Root-Core:
Hi there,

I've tested the music app and quite like it. There are some issues I encountered though.
One of them bothers me on an older laptop with an odd resolution (1366x768).

![Pre](https://user-images.githubusercontent.com/5329652/185978262-b712262a-34bb-4d73-92be-0dc153f356ff.png)

The seekbar is way to small to make actual use out of it, plus I have not found a way to seek forwards / backwards without it.
The arrow keys start the current track from beginning or skip to the next title. Long click on forward / backward does nothing (unlike the play button, which is on its own a bit counter intuitive to me and I use high playback speed all the time). It would be nice if the seekbar would span over the complete width of the player, as it sticks out of it anyways and therefore occupies the height.

_Random "artist impression":_

![Post](https://user-images.githubusercontent.com/5329652/185983205-53abf534-1b03-4ca0-be8e-f5e3dff9af15.png)

As there is a podcast functionality (which I use) you might keep in mind that one pixel could represent several minutes of content.

Feedback on this matter is appreciated. I could look into it and create a PR, but I don't know yet which implications such change would have on  mobile, as it seems there is no seekbar at all atm.

Edit:
// Okay, I checked the screenshots and the seekbar does not stick out on other resolutions. So I'm not sure if this would be a viable solution to the problem.
// The seekbar sticks out only if the details panel is shown, but even if it's closed it is rather hard to use on large files.

paulijar:
> Okay, I checked the screenshots and the seekbar does not stick out on other resolutions.

Yeah, you have really found an edge case here. The time indicator is not really supposed to get divided to two rows and the seekbar ought not to hang below the controls pane, but it does happen when all of these conditions are met:
- Details pane is open
- Window width is approximately between 1357 and 1368 pixels
- The length of the playing audio is more than one hour

I once used a lot of time in manually tweaking all the break points in window width where each UI element gets hidden, in attempt to prevent the UI from ever looking weird or broken like this. But the thing is, neither podcasts nor internet radio were yet supported at that time and I never tested with content length more than an hour.

> Long click on forward / backward does nothing

[...]

PR #1006, opened 2022-08-23
Enabled playback speed context menu via right click

Root-Core:
On the PC it just feels weird to long press something with the mouse, therefore I want to allow calling the context menu via right click.

I also replaced `getElementById()` with an event on the button element itself. Feel free to revert this change, if it was deliberate.

### DIFF
--- a/js/app/controllers/playercontroller.js
+++ b/js/app/controllers/playercontroller.js
@@ -413,15 +413,15 @@ function ($scope, $rootScope, playlistService, Audio, gettextCatalog, Restangula
 		$scope.playbackRate = stepRates[nextStep];
 	};
 
-	/** Context menu on long press of the play/pause button */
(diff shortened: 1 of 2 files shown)


paulijar:
Thanks, seems to work well. I was just thinking about adding this right-click option after reading your issue https://github.com/owncloud/music/issues/1004.

> I also replaced getElementById() with an event on the button element itself. Feel free to revert this change, if it was deliberate.

I actually didn't know of the `ngOn` directive and that you may use it to bind any custom events within the HTML template. But this makes the event binding more consistent, so I like it.

paulijar:
This PR has now been released in Music v1.7.0.

T0mWz:
> Where should this menu be visible? 🙈

Woow, found it finally. It's behind the play button, that is well hidden. Maybe an idea to add it as a separate button, like 'Playback speed', which opens this option?

PR #1007, opened 2022-08-24
Implemented a "timestamp preview" for the seek bar

Root-Core:
The timestamp that a click on the seek bar would set is shown in advance. This makes it easier to navigate in (not only) bigger files.

I also implemented touch support, which makes using the seekbar with touch way easier imho.
I have tested the touch part on a Microsoft Surface only yet, but it seems fine.

Mouse only: The preview is hidden after 1000ms while a track is played back or if the mouse moves out of the seek bar.

The feature is implemented into the normal player and in the file-plugin one.

### DIFF
--- a/js/app/controllers/playercontroller.js
+++ b/js/app/controllers/playercontroller.js
@@ -28,6 +28,7 @@ function ($scope, $rootScope, playlistService, Audio, gettextCatalog, Restangula
 	$scope.position = {
 		bufferPercent: '0%',
 		currentPercent: '0%',
+		currentPreview: null,
 		current: 0,
 		total: 0
 	};
@@ -519,6 +520,20 @@ function ($scope, $rootScope, playlistService, Audio, gettextCatalog, Restangula
 				setCurrentTrack(track);
 			}
 		}
+	};
+
+	$scope.seekbarPreview = function($event) {
+		if (!$scope.player.seekingSupported()) return;
+
+		var offsetX = $event.offsetX || $event.originalEvent.layerX;
+		var ratio = offsetX / $event.currentTarget.clientWidth;
+		var timestamp = ratio * $scope.position.total;
+		
+		$scope.position.currentPreview = timestamp;
+	};
+
+	$scope.seekbarLeave = function() {
+		$scope.position.currentPreview = null;
 	};
 
 	$scope.seekOffset = function(offset) {
--- a/js/app/controllers/playercontroller.js
+++ b/js/app/controllers/playercontroller.js
@@ -32,7 +32,7 @@ function ($scope, $rootScope, playlistService, Audio, gettextCatalog, Restangula
 		currentPreview: null,
 		currentPreview_ts: null, // Activation time stamp (Type: Date)
 		currentPreview_tf: null, // Transient mouse movement filter (Type: Number/Timer-Handle)
-		previewVisible: () => $scope.position.currentPreview && !$scope.position.currentPreview_tf,
+		previewVisible: () => ($scope.position.currentPreview !== null) && !$scope.position.currentPreview_tf,
 		current: 0,
 		total: 0
 	};
@@ -541,17 +541,22 @@ function ($scope, $rootScope, playlistService, Audio, gettextCatalog, Restangula
 	};
 
 	// Seekbar preview mouse support
-	$scope.seekbarPreview = function($event) {
-		if (!$scope.player.seekingSupported()) return;
-
+	function updateSeekPreview(xCoordinate) {
 		var seekBar = $('.seek-bar');
-		var offsetX = $event.clientX - seekBar.offset().left;
+		var offsetX = xCoordinate - seekBar.offset().left;
+		offsetX = Math.min(Math.max(0, offsetX), seekBar.width());
 		var ratio = offsetX / seekBar.width();
 		var timestamp = ratio * $scope.position.total;
-		
+
 		$scope.position.previewPercent = ratio * 100;
 		$scope.position.currentPreview = timestamp;
-		$scope.position.currentPreview_ts = Date.now();
+	}
+
+	$scope.seekbarPreview = function($event) {
+		if ($scope.player.seekingSupported()) {
+			updateSeekPreview($event.clientX);
+			$scope.position.currentPreview_ts = Date.now();
+		}
 	};
 
 	$scope.seekbarEnter = function() {
@@ -566,24 +571,16 @@ function ($scope, $rootScope, playlistService, Audio, gettextCatalog, Restangula
 
 	// Seekbar preview touch support
 	$scope.seekbarTouchLeave = function($event) {
-		if (!$scope.player.seekingSupported() || $event?.type !== 'touchend') return;
-		
-		// Reverse calculate on seek position
-		var ratio = $scope.position.currentPreview / $scope.position.total;
-		$scope.player.seek(ratio);
-		$scope.seekbarLeave();
+		if ($scope.player.seekingSupported() && $event?.type === 'touchend') {
+			$scope.player.seek($scope.position.previewPercent / 100);
+			$scope.seekbarLeave();
+		}
 	};
 
 	$scope.seekbarTouchPreview = function($event) {
-		if (!$scope.player.seekingSupported()) return;
-
-		var rect = $event.target.getBoundingClientRect();
-		var x = $event.targetTouches[0].clientX - rect.x;
-		var offsetX = Math.min(Math.max(0, x), rect.width);
-		var ratio = offsetX / rect.width;
-		var timestamp = ratio * $scope.position.total;
-
-		$scope.position.currentPreview = timestamp;
+		if ($scope.player.seekingSupported($event.targetTouches[0].clientX)) {
+			updateSeekPreview($event.targetTouches[0].clientX);
+		}
 	};
 
 	$scope.seekOffset = function(offset) {
--- a/js/app/controllers/playercontroller.js
+++ b/js/app/controllers/playercontroller.js
@@ -29,6 +29,7 @@ function ($scope, $rootScope, playlistService, Audio, gettextCatalog, Restangula
 		bufferPercent: '0%',
 		currentPercent: '0%',
 		currentPreview: null,
+		currentPreview_ts: null,
 		current: 0,
 		total: 0
 	};
@@ -108,6 +109,14 @@ function ($scope, $rootScope, playlistService, Audio, gettextCatalog, Restangula
 						onEnd();
 					}
 				}
+			}
+		}
+
+		// Show progress again instead of preview after a timeout of 1000ms
+		if ($scope.position.currentPreview_ts) {
+			var timeSincePreview = Date.now() - $scope.position.currentPreview_ts;
+			if (timeSincePreview >= 1000) {
+				$scope.seekbarLeave();
 			}
 		}
 	});
@@ -335,6 +344,7 @@ function ($scope, $rootScope, playlistService, Audio, gettextCatalog, Restangula
 			$scope.position.current = 0;
 			$scope.position.currentPercent = 0;
 			$scope.position.currentPreview = null;
+			$scope.position.currentPreview_ts = null;
 			$scope.position.bufferPercent = 0;
 			$scope.position.total = 0;
 		}
@@ -532,10 +542,12 @@ function ($scope, $rootScope, playlistService, Audio, gettextCatalog, Restangula
 		var timestamp = ratio * $scope.position.total;
 		
 		$scope.position.currentPreview = timestamp;
+		$scope.position.currentPreview_ts = Date.now();
 	};
 
 	$scope.seekbarLeave = function() {
 		$scope.position.currentPreview = null;
+		$scope.position.currentPreview_ts = null;
 	};
 
 	// Seekbar preview touch support
--- a/js/app/controllers/playercontroller.js
+++ b/js/app/controllers/playercontroller.js
@@ -26,11 +26,13 @@ function ($scope, $rootScope, playlistService, Audio, gettextCatalog, Restangula
 	$scope.shuffle = (localStorage.getItem('oc_music_shuffle') === 'true');
 	$scope.playbackRate = 1.0;  // rate can be 0.5~3.0
 	$scope.position = {
-		bufferPercent: '0%',
-		currentPercent: '0%',
+		bufferPercent: 0,
+		currentPercent: 0,
+		previewPercent: 0,
 		currentPreview: null,
 		currentPreview_ts: null, // Activation time stamp (Type: Date)
 		currentPreview_tf: null, // Transient mouse movement filter (Type: Number/Timer-Handle)
+		previewVisible: () => $scope.position.currentPreview && !$scope.position.currentPreview_tf,
 		current: 0,
 		total: 0
 	};
@@ -44,6 +46,9 @@ function ($scope, $rootScope, playlistService, Audio, gettextCatalog, Restangula
 	const PLAYBACK_RATE_STEPPING = 0.25;
 	const PLAYBACK_RATE_MIN = 0.5;
 	const PLAYBACK_RATE_MAX = 3.0;
+
+	$scope.min = Math.min;
+	$scope.abs = Math.abs;
 
 	// shuffle and repeat may be overridden with URL parameters
 	if ($location.search().shuffle !== undefined) {
@@ -408,11 +413,11 @@ function ($scope, $rootScope, playlistService, Audio, gettextCatalog, Restangula
 
 		$scope.position.current = position;
 		$scope.position.total = duration;
-		$scope.position.currentPercent = (duration > 0) ? position/duration*100 + '%' : 0;
+		$scope.position.currentPercent = (duration > 0) ? position/duration*100 : 0;
 	};
 
 	$scope.setBufferPercentage = function(percent) {
-		$scope.position.bufferPercent = Math.min(100, percent) + '%';
+		$scope.position.bufferPercent = Math.min(100, percent);
 	};
 
 	$scope.play = function() {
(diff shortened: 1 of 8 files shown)


paulijar:
Nice idea. However, after testing this, I feel that it is a tad too aggressive in how it replaces the current playback timestamp whenever the mouse cursor happens to pass over the progress bar.

I wonder, how difficult would it be to implement an alternative concept where the seek bar would support seeking with drag-and-drop and the preview would be shown only during the dragging.

How is the touch support supposed to work? When testing with Chrome and Firefox on Android, all I could see was that the predicted timestamp was shown briefly when long pressing some position on the seekbar.

Root-Core:
> Nice idea. However, after testing this, I feel that it is a tad too aggressive in how it replaces the current playback timestamp whenever the mouse cursor happens to pass over the progress bar.

I actually do not cross that often the progress bar without the reason of using it, so it has not bothered me too much. But to mitigate a rested cursor, i've implemented the timeout.

> I wonder, how difficult would it be to implement an alternative concept where the seek bar would support seeking with drag-and-drop and the preview would be shown only during the dragging.

I also tought about it, this is why I wanted some feedback first. I like this direct approach, but I can understand if it is too aggressive for others.

> How is the touch support supposed to work? When testing with Chrome and Firefox on Android, all I could see was that the predicted timestamp was shown briefly when long pressing some position on the seekbar.

Well.. I've tested it with Firefox on Windows. It works fine there.
After you touch into the seekbar it starts the preview feature and you can leave the seekbar with you finger afterwards and it will still work. On the end of the touch gesture, the selected timestamp is used to seek to.

Touch

It is hard to capture, as windows is a sometimes bad with showing the touch points.. but I hope you get the idea.

paulijar:
> But to mitigate a rested cursor, i've implemented the timeout.

Maybe it would help if there was a small delay also *before* showing the prediction. Something like 500ms could do the trick. That way, there would be no flickering when the mouse cursor just passes over the seekbar on its way to somewhere else.

Root-Core:
> Maybe it would help if there was a small delay also _before_ showing the prediction. Something like 500ms could do the trick. That way, there would be no flickering when the mouse cursor just passes over the seekbar on its way to somewhere else.

Aside from tooltips, I can't recall any UI element that exhibits such behavior. Do you know such UI elements?
It seems I can't find any while I'm hovering random elements in programs and websites on my computer.

In my experience, everything is instant when it comes to mouse input. There might be a additional animation or something, but some action will be taken immediately. If not, it feels sluggish quite quickly. 

A new tooltip kind of element could be added instead of replacing the timestamp, but to be honest I think this would feel more obtrusive.

paulijar:
Well, this *is* kind of tootip-like feature, so that's not a bad comparison. But there are also other, even more similar cases: when you hover over the icon of a running application on the taskbar of Windows, there is a small preview of the application window. However, this window doesn't appear immediately but with a delay of maybe half a second. With a delay this short, it still feels natural but prevents the action from triggering when the mouse cursor just passes by. And the open tabs in the top of the Chrome window work just the same. What is common to tooltips, these small popups, and the timestamp preview, is that they are all previews of what will happen if you click now.

Root-Core:
These examples _are_ tooltips, which are additional information - almost always implemented as "popups" - to an existing object in directly correlating positioning with the physical representation of it and / or the mouse, which is not really the case with this feature. Most objects do have a hover effect immediately and show the tooltip later (which is also the case for your examples), without this it really feels unnaturally. This might be a philosophical discussion though. As I said, I'm not sure how often a user would accidentally cross the seekbar in first place to be really bothered by this behavior. Other players are essentially doing the same thing, so I was under the impression this might work out.

One thing to consider is, that if I got the time to fix the bug with the touch input, it should show the same or similar behavior to the mouse input and a long press to activate this feature would feel wrong to me to be honest.

paulijar:
> These examples are tooltips, which are additional information

And so is this feature ;)

> in directly correlating positioning

Of course this feature is not *exactly* the same but still very similar. It's not really a significant difference if the thing appearing on hover is directly above the mouse cursor or two centimeters away. 

> Most objects do have a hover effect immediately and show the tooltip later

There is no hover effect on the seekbar and I wouldn't call this timestamp preview a hover effect either. But the mouse cursor does change when hovering over the seekbar and this kind of serves the same purpose as highlighting a thing which can be interacted with.

> I'm not sure how often a user would accidentally cross the seekbar in first place

It happens although not all the time. There are clickable things both on left and right side of the seek bar as well as below and above so sometimes the mouse cursor will cross the seek bar. And it will happen more often if we make the seek bar wider and sit below the other controls as discussed in https://github.com/owncloud/music/issues/1004.

> Other players are essentially doing the same thing

I take your word on that there are such players. But personally, I have never seen a feature like this on any player so it can't be all that common approach. Youtube, for instance, has a tooltip-like label following the mouse cursor and it appears with a small delay.

> Soundcloud for example uses a similar approach previewing the timestamp, it is also that direct.

I just visited https://soundcloud.com/ but I didn't see any feature like this.

> long press to activate this feature would feel wrong to me to be honest

Yes, obviously it's not a method which would be found or used by anyone. It was just the only way I could find to show the preview on touch devices when testing everything.

> I've tested it with Firefox on Windows. It works fine there.

Did you use some actual touch device to test this? Because I couldn't make it work using the touch emulation of Firefox developer tools on a non-touch device. Other than the Android phone, I have no touch devices to test with.

PS. Could you rebase your branch on top of the latest HEAD state of the master branch?

Root-Core:
> > These examples are tooltips, which are additional information
> 
> And so is this feature ;)

Hey! Do not end your quotes mid-sentence. :)

> Of course this feature is not _exactly_ the same but still very similar. It's not really a significant difference if the thing appearing on hover is directly above the mouse cursor or two centimeters away.

Well.. my other monitor is a 32:9 2xWQHD 49" one.. in that special case "two centimeters" is not really the appropriate term.

> There is no hover effect on the seekbar and I wouldn't call this timestamp preview a hover effect either. But the mouse cursor does change when hovering over the seekbar and this kind of serves the same purpose as highlighting a thing which can be interacted with.

It is no hover effect, that's true. But it is triggered by hovering. The mouse cursor change or some immediate indication is important, this is true.

> It happens although not all the time. There are clickable things both on left and right side of the seek bar as well as below and above so sometimes the mouse cursor will cross the seek bar. And it will happen more often if we make the seek bar wider and sit below the other controls as discussed in #1004.

If we redesign the controls, there are less UI elements "left and right of the seek bar" on the other hand. ;)
It is just some pixels tall, so the cursor passes the seek bar really fast. I might look if I could ignore these fast transients. A real delay feels bad in my eyes. Once you work with old computers for a moment, which are incredibly fast, smartphones feel sluggish again.

> > Other players are essentially doing the same thing
> 
> I take your word on that there are such players. But personally, I have never seen a feature like this on any player so it can't be all that common approach. Youtube, for instance, has a tooltip-like label following the mouse cursor and it appears with a small delay.

I had found some examples.. can't really remember them right now though. It's pretty late / early here right now.
Youtube has no delay for me, but it depends on the browser, monitor, refresh rate, os and computational power.. a little fade in seems apparent.

> I just visited https://soundcloud.com/ but I didn't see any feature like this.

Soundcloud

> Yes, obviously it's not a method which would be found or used by anyone. It was just the only way I could find to show the preview on touch devices when testing everything.

Sure, I just wanted to emphasize that coherence in the UX is important.

> Did you use some actual touch device to test this? Because I couldn't make it work using the touch emulation of Firefox developer tools on a non-touch device. Other than the Android phone, I have no touch devices to test with.

Yes, it was a Surface Pro with touch. The *.gif screenshot (in a previous post) is a recording of me touching on the screen.
Windows will most likely mix touch and mouse events.. MS is not necessarily known for sticking to standards.. but in this case it's not really clear if there are any standards in the first place.

> PS. Could you rebase your branch on top of the latest HEAD state of the master branch?

Sure thing, I'll do it while I fix the touch input.


PS. I think it's really interesting to discuss such topics as there are so many bad user interfaces out there, often from big companies, which could be fixed by little adjustments.

paulijar:
> It is just some pixels tall, so the cursor passes the seek bar really fast. I might look if I could ignore these fast transients. A real delay feels bad in my eyes.

The 500 ms I mentioned earlier was just an example. It's well possible, that something like 100-300 ms would be enough, but this would require some experimenting. With a short enough delay, the user might not even realize that there is any delay. 

On the other hand, the hide delay could maybe be longer than 1 second because now I feel that it may get hidden while I'm still considering if this is the right place to jump or not. Something like 2-3 secs could work.

> Youtube has no delay for me

I believe there is a delay but it is a really short one. Much shorter than on the Windows taskbar for instance. And I'm not certain if this is by design or just an implementation detail. But the preview does not flicker if the mouse cursor passes quickly over the seek bar, even if the hover effect of the seek bar would flicker.

Re: soundcloud

I guess the widget like the gif animation is shown only for the logged in users as there's no such thing when playing the songs on the landing page? I didn't care to create an account at this time.

That light orange coloring of the hover position seems like a good idea; I feel it makes it more clear that the shown timestamp is a prediction of the jump position. Maybe we should do something similar.

Root-Core:
> > It is just some pixels tall, so the cursor passes the seek bar really fast. I might look if I could ignore these fast transients. A real delay feels bad in my eyes.
> 
> The 500 ms I mentioned earlier was just an example. It's well possible, that something like 100-300 ms would be enough, but this would require some experimenting. With a short enough delay, the user might not even realize that there is any delay.

Yes, 500ms is way too much. I will play around with different delays, so that it feels right. The goal should clearly be that it doesn't feel like a delay at all.

> On the other hand, the hide delay could maybe be longer than 1 second because now I feel that it may get hidden while I'm still considering if this is the right place to jump or not. Something like 2-3 secs could work.

I actually do quite a lot of tests for such delays and less than 1000ms was way to fast. 1000ms feel okay as I'm able to remember where I parked my cursor (if I would move the cursor, the timestamp would change), but sure 1000ms is just the minimum and I don't mind raising the delay. It actually is a bit more than 1000ms already as the update rate of the player is not 100% accurate / it's resolution is not per ms.

> > Youtube has no delay for me
> 
> I believe there is a delay but it is a really short one. Much shorter than on the Windows taskbar for instance. And I'm not certain if this is by design or just an implementation detail. But the preview does not flicker if the mouse cursor passes quickly over the seek bar, even if the hover effect of the seek bar would flicker.

It's not really important tbh, it's fast and works. The up-sliding animation is sufficient enough to filter transients.

> > Re: soundcloud
> 
> I guess the widget like the gif animation is shown only for the logged in users as there's no such thing when playing the songs on the landing page? I didn't care to create an account at this time.

Nope. I'm not logged in. I was just playing back some track. It's the wave form on the page of a track.
I just clicked some random thing.. it seems you have to click the title, not the play button on the cover / thumbnail.

> That light orange coloring of the hover position seems like a good idea; I feel it makes it more clear that the shown timestamp is a prediction of the jump position. Maybe we should do something similar.

I have thought about that too. I was a bit intimidated to change the actual design of the seek bar in a first implementation. It has 3 different colors already - I think - and as it is still as small as it is, it seemed overwhelming if multiple layers showed up all at the same moment. It makes a whole lot more sense, if we actually redesign the seek bar.

It would be a good feature indeed. Any idea about the color scheme? It would be nice if there are accent colors defined by the theme or something. I have never looked into this stuff on OC/NC.

Root-Core:
I added the transient filter and changed the reset delay to ~2000ms. Git had some problems with the rebase it seems though. I could reset the branch and force push the current change if you want.

paulijar:
Thanks. I like the way it works now. And sorry for the delay, I have been busy with other things.

> It would be a good feature indeed. Any idea about the color scheme? It would be nice if there are accent colors defined by the theme or something. I have never looked into this stuff on OC/NC.

I don't believe that there would be any readily available accent color to use here. At least not a one which would be available on all supported cloud versions. The current progress bar has the color set as constant value here: https://github.com/owncloud/music/blob/fdba21e1aded8ff8333104694c011365d8997751/css/music-controls.css#L169
The buffered bar has the same color but with opacity 0.1. The preview effect could have the same color with opacity 0.5 or something.

js/app/controllers/playercontroller.js
```diff
@@ -479,10 +493,61 @@ function ($scope, $rootScope, playlistService, Audio, gettextCatalog, Restangula
...
+		var ratio = offsetX / $event.currentTarget.clientWidth;
+		var timestamp = ratio * $scope.position.total;
+		
+		$scope.position.currentPreview = timestamp;
+		$scope.position.currentPreview_ts = Date.now();
+	};
+
+	$scope.seekbarEnter = function() {
+		// Simple filter for transient mouse movements
+		$scope.position.currentPreview_tf = setTimeout(function() {
```
paulijar, line 510: This is unnecessarily convoluted. We can use the angular function $timeout, and the callback will be automatically executed within the angular digest loop. Something like
`$scope.position.currentPreview_tf = $timeout(() => $scope.position.currentPreview_tf = null, 100);`

js/app/controllers/playercontroller.js
```diff
@@ -479,10 +493,61 @@ function ($scope, $rootScope, playlistService, Audio, gettextCatalog, Restangula
...
+
+	$scope.seekbarLeave = function() {
+		$scope.position.currentPreview = null;
+		$scope.position.currentPreview_ts = null;
+	};
+
+	// Seekbar preview touch support
+	$scope.seekbarTouchLeave = function($event) {
+		if (!$scope.player.seekingSupported() || $event?.type !== 'touchend') return;
+
```
paulijar, line 526: I usually try to avoid multiple exit points in the functions, unless the additional exit point shortens or simplifies the function significantly. Here, inverting the if condition and having the logic within the body of the if wouldn't be any more complex:
```
if ($scope.player.seekingSupported() && $event?.type === 'touchend') {
    // do the calculation and seeking
}
```

Root-Core: I like to keep the functions as lean as possible. I worked on a big project from the callback **hell** once and have refactored the whole code base.. since then I try to avoid nesting and if something is considered as abort condition, I will do it fast and lean on the beginning of the function.

But no problem, I will change the corresponding locations.

js/app/controllers/playercontroller.js
Root-Core: Unfortunately, this does not work. Sure, I can make use of the $timeout function, but the scope does not update on changed object properties. You have to manually apply the changes or set watchers. I have tried to find a better solution, but this was the only working one without setting watchers etc.

However, it's been a few years since I last used my AngularJS voodoo.. so don't take my word for it.

Root-Core:
> Thanks. I like the way it works now. And sorry for the delay, I have been busy with other things.

No problem, I'm glad you like it.

> I don't believe that there would be any readily available accent color to use here. At least not a one which would be available on all supported cloud versions. The current progress bar has the color set as constant value here:
> 
> https://github.com/owncloud/music/blob/fdba21e1aded8ff8333104694c011365d8997751/css/music-controls.css#L169

Hm.. not sure if this will work, as you might want to seek back and if the progress bar is at 1.0 opacity, the preview would not show at all. We would have to calculate the color instead, which might be difficult for non grey-scale colors.

It also would be nice to see the current progress bar and the preview at the same time, which should be achievable with an overlaying, brighter color with opacity.

paulijar:
> Hm.. not sure if this will work, as you might want to seek back and if the progress bar is at 1.0 opacity, the preview would not show at all.

Yeah, it would require some additional logic at least. The way it works on Soundcloud is that basically the progress bar and the hover prediction bar swap colors when the hover position is to the left from the current progress position. So the darker bar is always on the left and the lighter bar to right from it. But what dark and light color indicate, depends on if the hover position is in the past or in the future. This sounds kind of complex, but in action, it feels quite intuitive.

js/app/controllers/playercontroller.js
paulijar: I just tested with my suggested line of code and it worked perfectly for me.

Root-Core:
I created a little mockup by cloning the progress bar and editing the style.. I just picked some color from the header and gave it 25% opacity. 

15% progress: grafik

75% progress: grafik

paulijar:
If you would now set in your mock-up the width of the `.play-bar` to be `min(currentPercent, previewPercent)`, then it should look similar to Souncloud, right?

Root-Core:
Weird.. it seems the screenshots got cropped after posting them. But you still can get the idea..

> If you would now set in your mock-up the width of the .play-bar to be min(currentPercent, previewPercent), then it should look similar to Souncloud, right?

I don't think so, they use a different approach. They seem to use the color for both with opacity for the preview. If the preview is lower than progress, they swap colors. In that case the preview would also need to set it's position accordingly to ```previewPercent``` and it's width to ```currentPercent - previewPercent```, if I'm not mistaken.

I'm not sure if the screenshots show the full behavior of the Soundcloud seek bar. I can not make them right now, sorry.

paulijar:
> They seem to use the same color for both with opacity for the preview. If the preview is less than progress, they swap colors.

Yes, this is one way to see it.

> In that case the preview would also need to set it's position accordingly to previewPercent and it's width to currentPercent - previewPercent, if I'm not mistaken.

Yes, I believe we have a common understanding of this. So if we define:
A = current playback percent
B = current hover prediction percent

Then we will get the Soundcloud-like behavior if we set
```
.opaque-bar {
    opacity: 1.0;
    left: 0;
    width: min(A, B)
}
.translucent-bar {
    opacity: 0.5;
    left: min(A, B);
    width: abs(A - B);
}
```
Obviously, that's not correct css syntax and we can't write such dynamic behavior into css file but I hope you got the idea.

Root-Core:
So we want to implement a Soundcloud-like behavior, right?
I think your pseudo code should work like that.

------------------

> Obviously, that's not correct css syntax and we can't write such dynamic behavior into css file but I hope you got the idea.

Are you *really* sure about this?

You actually can define variables in CSS and change them via JavaScript. This use case should absolutely be possible. The math operations ```min()```,  ```max()``` and  ```abs()``` are supported. Sure, you need to change the syntax a little.. but it should work. 

So the dynamic behavior is implemented in the CSS file, but triggered externally via JS.
Sorry for getting side tracked..

paulijar:
> So we want to implement a Soundcloud-like behavior, right?

Yes. I mean, some other system might work just as well. But in Soundcloud we have an example which looks and feels nice and intuitive, so copying it seems like an easy and safe option.

> Are you *really* sure about this?

Well, not really. It's possible that CSS can nowadays do all kinds of neat tricks which I'm not aware of. But one thing to consider here is again the IE support. All the neatest tricks are probably not supported on IE. It's okay if the eye-candy doesn't show up on IE but it's not okay if the current basic behavior gets broken. Luckily, ownCloud adds the class "ie" on the root <html> element when running on IE. Hence, IE-specific styles can be defined like this:
https://github.com/owncloud/music/blob/5c06f00c7fa21f6dfff39354c7be1e9828a551b2/css/music-controls.css#L96

paulijar:
This branch would need another rebasing. This time it will involve resolving some conflicts but nothing too severe, I believe.

paulijar:
Rebased

Root-Core:
I'm currently a bit busy.. I have rebased it locally and started working on it / refactoring the seekbar, but my time will be severly constrained the next weeks. I didn't want to generate noise with unfinished code.

paulijar:
I have now been working on the translucent preview bar a bit. I have implemented the first working version for the main app which just needs a bit of polishing and more testing. I haven't looked yet on the embedded Files player.

paulijar:
@Root-Core Do you want me to push my "preview bar" changes to your branch now, or would you like to finish your refactoring first?

paulijar:
@Root-Core Are you planning to work on this during the next few days? Because of the https://github.com/owncloud/music/issues/1017, I'd like to release Music v1.7.0 soon and it would be nice to get this PR included there. It's also an option to release this feature without the working touch support, and fix that in some later release. Or I could try to look into it myself although I'm not yet sure if I have time for that.

paulijar:
I now managed to fix the touch support as well as a few other small issues. Furthermore, I implemented the translucent preview bar also for the embedded Files player. I have pushed all my changes to the PR branch and rebased it again, making it ready for merging.

paulijar:
Merged. Thanks again for all your hard work on this!

paulijar:
This PR is now released in Music v1.7.0.

PR #1013, opened 2022-09-23
QoL improvements / hotkeys

Root-Core:
- [Reworked playback speed functionality](https://github.com/owncloud/music/commit/638cb5d14c4be31b3849caf39cd724bf591898b3)
- [QoL improvements / hotkeys](https://github.com/owncloud/music/commit/31c75bb5848bce1d6e16631d6fc9645d39825d9a)
- Added mute/volume toggle
- Added the volume off icon to files player
- Added .vscode folder to .gitignore
- Replaced deprecated "which" with "code"/"key" for hotkeys
- Changed long press delay to 500ms. as it is default on Android and iOS

Hotkeys added:
- Up: Increase volume (5 / 20 %)
- Down: Decrease volume (5 / 20 %)
- M: Mute toggle
- J: Seek forward (5 / 30 sec)
- L: Seek backwards (5 / 30 sec)
- K: Alias for Space (Play/Pause)
- Shift+Period / Shift+>: Increase playback speed
- Shift+Comma / Shift+<: Decrease playback speed

-------------------------

I tried to adopt the shortcuts of Youtube. I am a bit unhappy about the fact, that the arrow left/right keys are in use already, which differs from it. The focus here is more on music, so it might be the better configuration though.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,6 +19,9 @@ CVS/*
 .svn/*
 RCS/*
 
+# VSCode
+.vscode/
+
 # kdevelop
 .kdev
 *.kdev4
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,16 +1,30 @@
 ## [Unreleased]
 ### Added
+- Muting/unmuting by clicking the speaker icon
+  [#1013](https://github.com/owncloud/music/pull/1013) @Root-Core
+- Many new keyboard shortcuts
+  * Numpad +/-: Increase/decrease volume
+  * M: Mute toggle
+  * J/L: Seek backwards/forward
+  * K: Play/Pause toggle
+  * Shift + Comma/Period: Decrease/Increase playback speed
+  * Arrow Left/Right: Seek backwards/forward (was formerly skip previous/next)
+  * Ctrl + Arrow Left/Right: Skip previous/next
+  * Step size of seeking and volume control is increased when shift held down
+  [#1013](https://github.com/owncloud/music/pull/1013) @Root-Core
 
 ### Changed
 - Use background color definitions from the cloud core when available. Fixes a problem on Breeze Dark theme introduced in v1.6.0.
   [#1002](https://github.com/owncloud/music/pull/1002)
 - Subsonic: Search functions now find also songs by artist or album name and albums by artist name
   * This prevents the Substreamer client from going haywire when shuffle play for an artist requested (!)
-    [#1000](https://github.com/owncloud/music/issues/1000)
+  [#1000](https://github.com/owncloud/music/issues/1000)
 - Subsonic: Method `getCoverArt` returns a placeholder image (instead of an error) if the album/artis in question has no cover art set
   [#1000](https://github.com/owncloud/music/issues/1000)
 - Context menu on the play/pause button can be opened with right click in addition to the long press
   [#1006](https://github.com/owncloud/music/pull/1006) @Root-Core
+- Playback speed change by clicking the menu option now has step size 0.25 instead of 0.5. Right-click or long-press decreases the speed.
+  [#1013](https://github.com/owncloud/music/pull/1013) @Root-Core
 
 ### Fixed
 - Small issues in the mobile and tablet layouts
--- a/css/embedded/files-music-player.css
+++ b/css/embedded/files-music-player.css
@@ -259,8 +259,6 @@
 }
 
 #music-controls #volume-icon {
-	cursor: default;
-	opacity: .4;
 	position: absolute;
 	top: 0;
 	left: 0;
--- a/css/music-controls.css
+++ b/css/music-controls.css
@@ -198,8 +198,6 @@ body:not(.snapjs-left) #controls.taller-header {
 }
 
 #controls #volume-icon {
-	cursor: default;
-	opacity: .4;
 	position: absolute;
 	top: 0;
 	left: 0;
--- a/js/app/controllers/playercontroller.js
+++ b/js/app/controllers/playercontroller.js
@@ -352,6 +352,12 @@ function ($scope, $rootScope, playlistService, Audio, gettextCatalog, Restangula
 		$scope.player.setPlaybackRate(newValue);
 	});
 
+	$scope.offsetVolume = function (offset) {
+		const value = $scope.volume + offset;
+		$scope.volume = Math.max(0, Math.min(100, value)); 		// Clamp to 0-100
+		lastVolume = null;
+	};
+
 	$scope.toggleVolume = function() {
 		if (lastVolume) {
 			$scope.volume = lastVolume;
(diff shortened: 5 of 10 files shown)


paulijar:
Thanks again for your contribution. A few comments:
- The tooltip "press and hold for more" is now shown also over the play/pause context menu which makes little sense
- For some reason, holding shift down and right-clicking the stop button doesn't open the context menu on Firefox although it does work on Chrome
- In my opinion, the playback rate should still roll over from 3.0 to 0.5 with a left-click (and 0.5 to 3.0 with right-click). Especially on touch devices, people may have hard time to figure out, how to decrease the speed. At the same time, the slider may be difficult to use on touch.
- The translation directive can't actually handle the syntax `<span translate>{{ playing ? 'Pause' : 'Play'}}</span>` used in the context menu; the string is always shown in English. We need to use here `<span>{{ playing ? ('Pause' | translate) : ('Play' | translate) }}</span>`
- Using `code` to specify the hotkeys doesn't work on IE10 and IE11 which are still supported by the app. Some other old browsers probably have the same issue. IE is no longer supported by Nextcloud but at lest some of the ownCloud 10.x versions support IE10+. Maybe it would make sense to have a polyfill for the keyboard events to enable use of the non-deprecated interface everywhere.
- We could maybe map the arrows left/right to seeking and then have the skip previous/next functionality on CTRL + left/right. At the time being, skipping to next track may happen a bit too easily by accident, considering that it is kind of destructive action: you lose the playback location where you were with no easy way to go back there.
- The arrows up/down already have the functionality of scrolling up/down so I wouldn't put the volume control there. What about mapping that to the + and - keys, including the ones on the number pad.
- With this many keyboard shortcuts, we should probably have some kind of hint somewhere listing all of them for the user. But that doesn't have to happen within this PR.

Root-Core:
>The tooltip "press and hold for more" is now shown also over the play/pause context menu which makes little sense

You are right. Should be an easy fix.
If I'm right, it was also shown if the context menu is already open (previous to my changes), which does make little sense to me.

So my solution would be to deactivate the tooltip altogether if the context menu is shown.

>For some reason, holding shift down and right-clicking the stop button doesn't open the context menu on Firefox although it does work on Chrome

This is strange, it opens the context menu of Firefox and the long press seems to kick in after the 500ms delay which opens our context menu. This could be a tricky one.

>In my opinion, the playback rate should still roll over from 3.0 to 0.5 with a left-click (and 0.5 to 3.0 with right-click). Especially on touch devices, people may have hard time to figure out, how to decrease the speed. At the same time, the slider may be difficult to use on touch.

Sure thing, I implemented multiple approaches of the function and missed the rollover in the final one.

>The translation directive can't actually handle the syntax `<span translate>{{ playing ? 'Pause' : 'Play'}}</span>` used in the context menu; the string is always shown in English. We need to use here `<span>{{ playing ? ('Pause' | translate) : ('Play' | translate) }}</span>`

I should have checked this. I already suspected something like this.

>Using `code` to specify the hotkeys doesn't work on IE10 and IE11 which are still supported by the app. Some other old browsers probably have the same issue. IE is no longer supported by Nextcloud but at lest some of the ownCloud 10.x versions support IE10+. Maybe it would make sense to have a polyfill for the keyboard events to enable use of the non-deprecated interface everywhere.

Hm.. IE is not my favorite browser, but fair enough if it is supported.
I actually wasn't sure about if it is a good idea to replace it, but as I do use a german keyboard layout and have tested the shortcuts with multiple layouts, code seemed the most sufficient. I didn't want to add some random dependency in form of a polyfill though.

Keyboard input is a really complex matter, I'm not sure how good a polyfill will fill the gap tbh.

I found this one: https://inexorabletash.github.io/polyfill/keyboard.html
The developer has archived the github repo for unknown reasons though.

>We could maybe map the arrows left/right to seeking and then have the skip previous/next functionality on CTRL + left/right. At the time being, skipping to next track may happen a bit too easily by accident, considering that it is kind of destructive action: you lose the playback location where you were with no easy way to go back there.

True, I didn't want to touch existing shortcuts without getting feedback first.

>The arrows up/down already have the functionality of scrolling up/down so I wouldn't put the volume control there. What about mapping that to the + and - keys, including the ones on the number pad.

In my dev instance are just a couple files, so I totally missed that one. Youtube activates the shortcuts only if the player is focused. This is not really a thing here, so this might be a viable solution. 

>With this many keyboard shortcuts, we should probably have some kind of hint somewhere listing all of them for the user. But that doesn't have to happen within this PR.

Absolutely.

paulijar:
> For some reason, holding shift down and right-clicking the stop button doesn't open the context menu on Firefox although it does work on Chrome

Seems like this is an intentional feature on Firefox: https://bugzilla.mozilla.org/show_bug.cgi?id=692139
That is, Firefox wants to provide a way to open the browser context menu also on those elements which have a context menu defined by the web page. And holding shift pressed is the mechanism they have selected for this.

I guess we can live with this. However, there is now also such a problem that the "shift mode" gets stuck on this shift + right-click on Firefox. That is, the stop button remains visible in the controls pane until when the shift button is pressed and released again. Maybe something could be done for this.

paulijar:
> I found this one: https://inexorabletash.github.io/polyfill/keyboard.html

Yes, I found the same project when i googled this yesterday. I did a quick test with it, and it seems to work fine in our case:
I downloaded https://github.com/inexorabletash/polyfill/blob/master/keyboard.js and placed it under /js/vendor/polyfill and required it in js/index.app.js. With this, the keyboard shortcuts started to work also on IE.

The only problem I saw was that the shift key handling didn't work on IE. This was because the polyfill set the `e.code` as 'Shift' instead of 'LeftShift' or 'RightShift'. Adding the `case 'Shift'` under the `switch` solved this issue.

Root-Core:
> Seems like this is an intentional feature on Firefox: https://bugzilla.mozilla.org/show_bug.cgi?id=692139 That is, Firefox wants to provide a way to open the browser context menu also on those elements which have a context menu defined by the web page. And holding shift pressed is the mechanism they have selected for this.

Yes, I remember that there was something like this. It is rather unlikely that this will happen too often I think, as in why should I hold shift and do a right click on that element. It's annoying though.

> With this, the keyboard shortcuts started to work also on IE.

It's less about that it will kind of work, but more about the way it works. It will differ from the native API as in it will not represent the physical key. (I think.)

>The only problem I saw was that the shift key handling didn't work on IE. This was because the polyfill set the e.code as 'Shift' instead of 'LeftShift' or 'RightShift'. Adding the case 'Shift' under the switch solved this issue.

This is because it does not represent the physical key, but it might work as a workaround.

paulijar:
> It's less about that it will kind of work, but more about the way it works. It will differ from the native API as in it will not represent the physical key. (I think.)

Yeah, it's not a perfect system, but it may be good enough for this case. I'm not totally convinced that keeping the IE compatibility is even worth the effort today but so far I have continued it with a "best effort" practice: I try to keep the core functionality intact and fix any IE-specific issues which are easy to fix. There remain some annoying graphical glitches on IE but I have decided not to touch them as they do not block any functionality and fixing them could be painful. So overall, I think that "mostly works well" is good enough target level for IE.

Root-Core:
If you want, you should be able to push your changes to my branch.
I thought about adding just a thin mapping of the used keys on IE, but that polyfill isn't that big either.

paulijar:
Seems that the pushing to your branch worked. Did you mark me as a collaborator on your clone or why was I able to do that?

paulijar:
I'm ready to merge this now, unless you still want to tweak something.

Root-Core:
> Seems that the pushing to your branch worked. Did you mark me as a collaborator on your clone or why was I able to do that?

If you open a pull request, you can grant write access to the according branch to people with write access to the repo you want to push to. This checkbox is checked by default I think.

grafik

------------------

> I'm ready to merge this now, unless you still want to tweak something.

I think it's ready. The only thing that bothers me is that IE thing, as some shortcuts could be problematic for non-US keyboard layouts. Solving this would be a major undertaking and not worth the effort though.

paulijar:
This PR is not released in Music v1.7.0.

Issue #1052, opened 2023-03-09
Shuffle through All tracks seems to be playing the same songs


rstefko:
How is the mechanisms implemented? Is it on client side or server side?

paulijar:
The shuffling happens on the client-side. Can you be a bit more specific on how have you observed this issue?

rstefko:
I have almost 4000 songs in the All tracks list. I use ownCloud Music as app in Edge. I play the music every work day. A lot of songs are very "liked" by the player, so it plays them every day. But there is a lot of songs that are never played.

It looks like "shuffle" generates almost the same numbers.

paulijar:
Okay, thanks for the explanation.

Deep down, the shuffling happens here using the lodash library:
https://github.com/owncloud/music/blob/d8590b97728558f594c13ec5e1ccba22bae96d06/js/app/services/playlistservice.js#L25

Now, according to [this](https://stackoverflow.com/questions/44678339/does-lodash-shuffle-provides-even-distribution), the lodash should create an even distribution. But maybe that is wrong, who knows.

It's also quite difficult to evaluate the randomness in day-to-day life because the sample size will inevitably be so small: If you randomly pick a song 40000 times from your collection of 4000 songs, then each song will be picked 10 times on average. But there is a high likelihood that some songs will happen to be picked much more often and some others less often. Because that's the nature of randomness. But if you pick a song one million times, then you would probably already get quite even distribution, provided that the shuffle works as it should. But picking even the 40000 songs takes probably more than 100 days of non-stop listening.

rstefko:
Thank you for explanation. I have created a new view called "Random" as alternative for the random button inside play navigator. It picks 300 random tracks. You see what is inside the list and can expect what songs will be played. Also you can refresh the page to get different set. Similar functionality is in the play:Sub application.

It is based on the All tracks view. It may share a lot of logic with the all tracks controller, but currently it is just a copy.

Please check the pull request and merge it if possible.

rstefko:
Great enhancement!

PR #1061, opened 2023-04-02
Added Random view

rstefko:
See #1052 for details.


paulijar:
Thanks for the PR and sorry for the long delay in giving any feedback. I tested this briefly a week ago but didn't have the time to write my impressions. Here are some thoughts:
1. I can see that this gives at least some added value. However, more work is needed before I could consider merging it.
2. The thing bothering me the most is how the view always re-randomizes its content upon view switch. This is not in line with the rest of the app and not very good user experience: view navigation back and forth should not be a destructive operation. An idea: maybe the first randomization could happen when the view is first entered and a new randomization would be available via a context menu item on the navigation pane.
3. The *All tracks* view has lot of complexity which would be totally unnecessary here. The whole "bucket" concept and using the `in-view-observer` directive have been invented to be able to handle lists of tens of thousands of tracks. When there can be only 300 tracks in the view, the only thing needed is a single `track-list` directive.
4. In the Subsonic API and at least on the DSub client, the "random songs" feature allows the user to optionally define start year, end year, and/or genre. Having that kind of ability here would bring some genuine added value and make it more obvious, why this view should exist in addition to the shuffle feature.
5. Please rebase your PR. I will not merge a complex mess of merge commits, I like my history nice and linear. Also, it would probably make your life easier to make the PR out of a separate feature branch instead of using the master branch of your fork for this.

paulijar:
@rstefko Are you planning to work further on this feature? It's also possible that I will pick up from where you left and do some finalization and polishing. But I'm not certain when would that happen and don't want to step on your toes. There will probably be an app release v1.9.0 during or after July with some new features, and also this one could maybe be released there.

rstefko:
@paulijar I am still testing it. I have pushed my latest changes. It looks like the sort by play count is finally fixing the initial problem #1052. It would be great if you could finish it. Thanks Roman

PR #1079, opened 2023-08-10
Add nop getNowPlaying route for SubsonicController

NattyNarwhal:
Subsonic clients call this API to see what media users are playing. If a client calls this, returning an empty list is better than an a 404 or code 70 for non-supported route, since what a server may return for unimplemented API routes may vary.

Ideally, this should be expanded to actually return what's actually playing on the server.

For an example of an affected client, see SubmarinerApp/Submariner#148.

### DIFF
--- a/lib/Controller/SubsonicController.php
+++ b/lib/Controller/SubsonicController.php
@@ -992,7 +992,7 @@ class SubsonicController extends Controller {
 	 */
 	protected function getNowPlaying() {
 		// TODO: not supported yet
-		return $this->subsonicResponse(['nowPlaying' => []]);
+		return $this->subsonicResponse(['nowPlaying' => ['entry' => []]]);
 	}
 
 	/* -------------------------------------------------------------------------
--- a/lib/Controller/SubsonicController.php
+++ b/lib/Controller/SubsonicController.php
@@ -987,6 +987,14 @@ class SubsonicController extends Controller {
 		]]);
 	}
 
+	/**
+	 * @SubsonicAPI
+	 */
+	protected function getNowPlaying() {
+		// TODO: not supported yet
+		return $this->subsonicResponse(['nowPlaying' => []]);
+	}
+
 	/* -------------------------------------------------------------------------
 	 * Helper methods
 	 *------------------------------------------------------------------------*/


paulijar:
Thanks!
As a small detail, when the response is defined like
```
$this->subsonicResponse(['nowPlaying' => []])
```
...the result will look like this in JSON and XML formats:
image image

It might make more sense to define the response as
```
$this->subsonicResponse(['nowPlaying' => ['entry' => []]])
```
...which would make it look like this:
image image

As a side note, it might actually make more sense for also us to return 404 on unimplemented endpoints. That seems to be what the original Subsonic server does on such situations (like on http://demo.subsonic.org/rest/foobar?u=guest&p=guest&c=test&v=1.16&f=json).

NattyNarwhal:
Second approach looks better XML wise, but the JSON looks likelier to confuse a client? At least Submariner doesn't care; it'll get the response, never hit any nowPlaying/entry XML elements, and finish parsing the response.

FWIW, Navidrome spits this out for an empty getNowPlaying:

```xml
<subsonic-response xmlns="http://subsonic.org/restapi" status="ok" version="1.16.1" type="navidrome" serverVersion="0.40.0-FREEBSD (8799358)"><nowPlaying></nowPlaying></subsonic-response>
```

```json
{"subsonic-response":{"status":"ok","version":"1.16.1","type":"navidrome","serverVersion":"0.40.0-FREEBSD (8799358)","nowPlaying":{}}}
```

I agree 404 is probably the right thing here. A Subsonic code 70 seems used more for resource not found, not API not implemented.

paulijar:
> JSON looks likelier to confuse a client? At least Submariner doesn't care; it'll get the response, never hit any nowPlaying/entry XML elements, and finish parsing the response.

IMHO, the structure of the JSON conveys the meaning "nothing currently played" quite well: There could potentially be multiple entries which is why the value of the key `entry` is an array, and now just the array is empty.

We can't easily test the case of an empty nowPlaying list on the public Subsonic demo server, but testing with another endpoint returning an empty list, the result seems to be pretty much analogous to what you have shown above for Navidrome:
http://demo.subsonic.org/rest/getSongsByGenre?u=guest&p=guest&c=test&v=1.16&f=json&genre=nosuchthing
http://demo.subsonic.org/rest/getSongsByGenre?u=guest&p=guest&c=test&v=1.16&f=xml&genre=nosuchthing
So maybe that would be the way to go. In our framework, this can be achieved with the (little hacky) syntax
```
$this->subsonicResponse(['nowPlaying' => new \stdClass]);
```

From practical point of view, it probably doesn't make much difference in case Submariner doesn't care. This is the first time I hear of any client actually using this part of the API, so it's well possible that there are no others. I don't have any Apple devices so I haven't personally tested Submariner or any of the other few available Apple clients.

NattyNarwhal:
> IMHO, the structure of the JSON conveys the meaning "nothing currently played" quite well: There could potentially be multiple entries which is why the value of the key entry is an array, and now just the array is empty.

This makes sense, I think going the first way you mentioned earlier is preferable in that light. If there's some pickier API consumer, it might be worth changing it to `\stdClass`.

I'm using the Cocoa XML parsing stuff and parsing an element at a time, so I don't think it cares too much. It's also quite an old codebase, and I'm surprised to see what corners of the Subsonic API it uses (...and doesn't).

paulijar:
This is now released as part of Music v1.9.1.

Issue #1080, opened 2023-08-13
EXCEPTION


Justman10000:
Hi, got this error:

> OC\IntegrityCheck\Exceptions\MissingSignatureException
> Signature data not found.

paulijar:
So what were you doing when you got this?

Justman10000:
> So what were you doing when you got this?

```bash
mkdir apps/music
cd apps/music
git init
git remote add origin https://github.com/owncloud/music.git
git pull origin master
git branch -m master
rm -r .* *.md COPYING LICENSE
```

Justman10000:
> So what were you doing when you got this?

Have installed this so:

```bash
mkdir apps/music
cd apps/music
git init
git remote add origin https://github.com/owncloud/music.git
git pull origin master
git branch -m master
rm -r .* *.md COPYING LICENSE
```

paulijar:
Are you planning to do development on this app, or do you want to use bleeding edge software which hasn't been released yet? Because for most users, using the pre-packaged deliveries should be the preferred option of installing. So either install it from the app store or download the tar.gz from https://github.com/owncloud/music/releases and extract it to the `apps` folder.

If you pull the project yourself from github, then obviously the application will not be signed and I assume that your error message is now somehow related to that. However, not having signature shouldn't cause any other problems except for the nagging on the admin UI.

Also, when pulling the project from github, you need to build the front-end bundle by running these commands under `apps/music`:
```
npm install --deps
npm run build
```
And for that, you need to have `npm` installed.

Justman10000:
> If you pull the project yourself from github, then obviously the application will not be signed
Can I sign this myself?

Justman10000:
> If you pull the project yourself from github, then obviously the application will not be signed

Can I sign this myself?

paulijar:
I don't believe there is any self-signing system, at least officially. If I remember correctly, I obtained the certificates used for signing years ago by registering to some services maintained by ownCloud and Nextcloud. I think the system was such that one application ID may be registered only once. That is, the application ID `music` is now permanently bound to a private key in my possession, and only I can create signed deliveries of this app. Hence, whenever a user sees a signed `music` application delivery, she can be reasonably certain that it is from me and not from some (possibly malicious) 3rd party.

But personally, I'm using an unsigned version of Music app 99% of the time. I only sign the app whenever I'm making a release. It would be cumbersome to sign it after each code change.

Justman10000:
[...]

PR #1082, opened 2023-08-14
Update

Justman10000:


Justman10000:
Well, having the latest of everything can never hurt

Justman10000:
Sure, you have to adapt the source code... Reasons for this are these permament warnings of outdated and/or abandoned major versions or even whole modules

Justman10000:
Sure, one has to adapt the source code... Reasons for this are these permament warnings of outdated and/or abandoned major versions or even whole modules

paulijar:
> Well, having the latest of everything can never hurt

Sure it can, and this PR just clearly demonstrated that.

> Sure, one has to adapt the source code... 

So feel free to submit a PR where you include also those mandatory source code changes. Any half-assed attempt to update the dependencies, breaking everything along the way, does not benefit anyone.

> Reasons for this are these permament warnings of outdated and/or abandoned major versions or even whole modules

So did the updates proposed here remove any of those warnings? I *think* that most of the warnings are related either directly or indirectly to AngularJS, which as stated elsewhere, would be a huge struggle to replace.

Justman10000:
To be able to test something in general, I would need to be able to use the plugin via the source code

paulijar:
Obviously, and it will be rather pointless to create any PRs before you have the ability to test them. Have you tested what I suggested in https://github.com/owncloud/music/issues/1080#issuecomment-1677855789? Also, when you have the signature problem, can you find any additional clues from owncloud.log or nextcloud.log?

PR #1094, opened 2023-10-04
Add rounded corners to album art

StratusFearMe21:
It's a simple change, but I think it looks so much better this way. Plus it matches the look of modern Nextcloud

### DIFF
--- a/css/music-albums.css
+++ b/css/music-albums.css
@@ -73,6 +73,7 @@
 	line-height: 190px;
 	font-size: 105px;
 	cursor: pointer;
+	border-radius: 18px
 }
 
 .album-area > .overlay {
--- a/css/music.css
+++ b/css/music.css
@@ -104,7 +104,7 @@ h2 span:hover .play {
 	background-size: cover;
 	background-position: center;
 	background-repeat: no-repeat;
-	border-radius: 18px;
+	border-radius: 12px;
 }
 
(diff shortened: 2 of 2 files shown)


paulijar:
Thanks for the PR.

I can see your point on matching the modern NC layout. However, after using this for a while, I got the feeling that this is a bit too much. I feel that I got an overdose of roundness if that makes any sense.

Also, as most album cover arts have been originally designed in square format, some of them don't look quite good when the corners are aggressively rounded. Especially, if the cover design includes some kind of narrow border on the edges, then it tends to look ugly as rounded. Here are a few examples:
image image image image

It might also make sense to follow the lead of the of the Files app in Nextcloud. There, the border-radius has been defined as `var(--border-radius)` which equals just 3px and the roundness is almost invisible:

image

The Files app uses the same radius also for the small icons in the list layout:
image

paulijar:
I now added the minuscule rounding of 3px on all album art. The implementation is such that if you prefer some other amount of rounding, it's easy to override this using the [Custom CSS](https://apps.nextcloud.com/apps/theming_customcss) app. To use the radius suggested in this PR, you may define the following custom CSS rule:
```
:root {
    --music-albumart-small-border-radius: 12px;
    --music-albumart-large-border-radius: 18px;
}
```

Issue #1123, opened 2024-01-31
Subsonic: wrong API response on `albumInfo2` API


perillamint:
Currently, Owncloud Music returns `albumInfo2` object with the API

https://github.com/owncloud/music/blob/master/lib/Controller/SubsonicController.php#L1524
https://github.com/owncloud/music/blob/master/lib/Controller/SubsonicController.php#L327

However, it is not a correct response. As per OpenSubsonic and Subsonic API spec

* https://opensubsonic.netlify.app/docs/endpoints/getalbuminfo2/
* https://www.subsonic.org/pages/api.jsp#getAlbumInfo2

It should return `albumInfo` object instead of `albumInfo2`.

paulijar:
Thanks for the report. You are right, `getAlbumInfo2` really should return `albumInfo` instead of `albumInfo2`. Meanwhile, `getArtistInfo2` returns `artistInfo2` and `getAlbumList2` returns `albumList2`. Probably I had just assumed that `getAlbumInfo2` would follow a similar pattern, without looking at the specification closely enough.

We can also see from the demo server, that the original Subsonic server is implemented as specified and this is not just a typo in the spec: http://demo.subsonic.org/rest/getAlbumInfo2?u=guest&p=guest&c=test&v=1.16&f=xml&id=22

paulijar:
The fix is now released in Music v1.11.0.

PR #1124, opened 2024-01-31
Subsonic: return `albumList` instead of `albumList2`

perillamint:
This PR fixes issue https://github.com/owncloud/music/issues/1123


paulijar:
Your assessment in https://github.com/owncloud/music/issues/1123 is correct but I think you accidentally modified the wrong method: You were supposed to modify `getAlbumInfo2`, not `getAlbumList2`.

perillamint:
Oh, I'll re-create PR then. My mistake. Thanks! @paulijar

perillamint:
Superseded by #1125

PR #1125, opened 2024-02-01
Subsonic: return `albumInfo` instead of `albumInfo2`

perillamint:
This PR fixes issue https://github.com/owncloud/music/issues/1123


paulijar:
Thanks. After this change, the first parameter of `doGetAlbumInfo` is kind of pointless as it's only ever given the value 'albumInfo'. Could you also refactor the code by removing this parameter and hard-coding the element name as 'albumInfo' within the body of `doGetAlbumInfo`?

PR #1141, opened 2024-04-07
Advanced search view

paulijar:
An advanced search view has been added to the web UI. This brings this powerful feature previously implemented only in the Ampache API to be available for a broader audience. The feature allows the user to make search queries where an arbitrary number of diverse criteria are combined. The link to this feature can be found from the context menu of the "Search" navigation item.

While implementing this, several bugs were found from the back-end logic of the advanced search. These were fixed which benefits also the Ampache API.

Also, some of the features of the advanced search were previously not supported when the cloud used SQLite or PostgreSQL as the database system. With this PR, the limitations have been lifted:
- Regular expressions can be used also with SQLite
- "Sounds like" type rules can be used also with SQLite and PostgreSQL. With the latter, the extension [fuzzystrmatch](https://www.postgresql.org/docs/current/fuzzystrmatch.html) needs to be installed.
- Searching by `album_genre` and `artist_genre` is supported also with PostgreSQL

image

Entity types which can be searched for:
image

Comparison operators available for string type properties:
image

All searchable properties available for tracks:
image


paulijar:
The feature is now released in Music v1.11.0.

djtulan:
I have a question about the advanced search feature.

Maybe it would be possible to add some more search terms?
But I'm not sure if the information is available for the advanced search.

What I really miss is the search for **comment** tag.
I prefer the mark my files with special comments, if I could search for it in the music app, that would be awsome.

I nice to have would be the search for **bpm** tag and **publisher** tag.

paulijar:
The advanced search can search only for details which Music app has stored in the database during the scanning. Searching for other details would be almost as slow as the initial scan of the music collection. It would also be technically very different from the DB search and require implementing a significant amount of additional logic.

All of the mentioned details are such that they are not currently stored in the DB. It would be possible to change the database schema to include these fields but I'm a bit reluctant to do that because
a) The DB schema changes are error prone and difficult to test comprehensibly. History has shown that about 50% of the times I end up screwing something up in the DB migration and at least some users will face problems in updating to the new Music app version.
b) The users would be forced to rescan their library to have these new fields scanned and stored.

The set of details currently supported has been selected as an intersection of what is stored in the DB and what is specified in the Ampache API as supported fields. The feature on the web UI is based on the same back-end logic implemented for the Ampache API. For tracks, the only search terms not currently supported, which we could easily add, would be the mime type and the disk number.

The reasons given above don't mean that I would never add any additional search terms. Just that it's not a small task and may not happen any time soon. If there will ever be other reasons to change the DB schema, then it might be worthwhile to add also these fields on the same run. Another song detail, which could interest some users based on the feature requests from the past, would be the composer.

paulijar:
> I have a question about the advanced search feature.
> 
> Maybe it would be possible to add some more search terms? But I'm not sure if the information is available for the advanced search.
> 
> What I really miss is the search for **comment** tag. I prefer the mark my files with special comments, if I could search for it in the music app, that would be awsome.
> 
> I nice to have would be the search for **bpm** tag and **publisher** tag.

@djtulan The new Music v3.2.0 now supports advanced search by comment, BPM, and publisher (record label). The library has to be rescanned to make these work.

djtulan:
> > I have a question about the advanced search feature.
> > Maybe it would be possible to add some more search terms? But I'm not sure if the information is available for the advanced search.
> > What I really miss is the search for **comment** tag. I prefer the mark my files with special comments, if I could search for it in the music app, that would be awsome.
> > I nice to have would be the search for **bpm** tag and **publisher** tag.
> 
> @djtulan The new Music v3.2.0 now supports advanced search by comment, BPM, and publisher (record label). The library has to be rescanned to make these work.

@paulijar Thank you very much for that extension, I will give it a try now!

EDIT: Works very well after a rescan.

PR #1158, opened 2024-07-13
Fixed plural translatable strings

flozz:
When working on the French translation of this app I found some plural strings that are not properly handled.

For example:

* `1 result`
* `{{ resultCount() }} results`

It works in English because `1` is singular and `0` or `>1` are plural. It does not work for some other languages like French where `0` and `1` are singular and `>1` is plural. So the advanced search displays "1 result" when there are no results.

To fix this, I changed the previous strings to:

* `{{ resultCount() }} result`
* `{{ resultCount() }} results`

It does not change anything for English, but now it works for French (and probably for some other languages too). :)


paulijar:
Thanks, makes sense. I updated the Finnish, German, and Chinese translations to match this change. The Finnish translation is the only completely up-to-date one while German and Chinese are partially up-to-date, having been updated around three years ago. The rest of the translations are totally out of date and probably don't even have any translations for the affected strings.

PR #1196, opened 2025-01-22
Use smart_ prefix

rstefko:
According to Ampache Docs (https://ampache.org/api/browse/playlist-browse/), Smartlists should be prefixed with "smart_" prefix. It makes the list being displayed eg. in Ample client.

### DIFF
--- a/lib/Controller/AmpacheController.php
+++ b/lib/Controller/AmpacheController.php
@@ -103,7 +103,7 @@ class AmpacheController extends ApiController {
 	private ?AmpacheSession $session;
 	private array $namePrefixes;
 
-	const ALL_TRACKS_PLAYLIST_ID = -1;
+	const ALL_TRACKS_PLAYLIST_ID = "smart_1";
 	const API4_VERSION = '4.4.0';
 	const API5_VERSION = '5.6.0';
 	const API6_VERSION = '6.6.1';


paulijar:
Yeah, I have noticed that Ample relies on this convention although I haven't seen this documented in the Ampache API specification before; there seems to be some new pages there since the last time I checked. Thanks for pointing that out.

However, the Music app currently assumes everywhere that the IDs used in the Ampache API would be integers. That's why Scrutinizer is nagging now. Quite a bit of more work than just changing the ID would be needed to properly support this. It would require some investigation and thinking to find out, if the scope of the type change can be limited to the playlists, or does this spread to everywhere in the Ampache API. So far, I haven't been bothered to check this. If you have time and interest to do this work, the help is appreciated.

rstefko:
Thanks. I have made some more research and this change is really not necessary. Newer version of https://www.npmjs.com/package/javascript-ampache library used in Ample already supports userSmartlists API metod.

I am working on a fork of Ample (based on version 2) and want to make it optimized for ownCloud Music. It can be accessed at [ample.fm](https://ample.fm). One of the nice features there is Smartlist Auto Play, which automatically adds tracks into the queue when it reaches the end. So I am looking at how to make it work.

Issue #1181, opened 2024-11-29
problems with ample setup


djtulan:
Hi!

I'm not sure if I'm right at this place, but I would like to ask for help here.
I'm running nextcloud as a snap image also with the music plugin installed. Everything works fine!
At the same server, but as a different instance, I run an apache2 server configured to run **ample** & **foam**.

see https://github.com/Andrew-McGee/foam
see https://github.com/mitchray/ample

I can run foam without any issues, but I don't managed to run ample.
Since ample looks more developed, I would prefer it to foam.

But I always run into troubles with CORS.

> Status
> 405
> Method Not Allowed
> VersionHTTP/1.1
> Übertragen1,26 kB (0 B Größe)
> Referrer Policystrict-origin-when-cross-origin
> DNS-AuflösungSystem

Since I cannot modify the snap image so easily, I would like to ask if you know any solution?
Maybe I do something completely wrong?

paulijar:
I believe that Foam makes the calls to the Ampache API from its back-end while Ample does them directly from its front-end. Because of this, Ample is affected by CORS restrictions but Foam is not. 

That is, as-is, Ample works only if the server (Nextcloud Music in this case) is served from the same domain as Ample itself. I think that the Ample developer said somewhere on his GitHub that if anyone wants to use the app in different kind of configuration, then it's up to the user to solve the CORS issues. I'm not a subject matter expert on this, but probably it would mean modifying some HTTP headers of Ample. Maybe this could be done in the Apache configuration or then it would require modification on the Ample code.

djtulan:
As I understand, the Ampache server must set that special headers
see: https://github.com/mitchray/ample/wiki/Connecting-to-a-remote-Ampache-server-(CORS)

I don't know if this would be possible inside the Music Plugin Ampache code?
If so, I could try to set it in the plugin code.

paulijar:
I managed to enable the cross-origin calls to the Ampache API. It turned out that there was already good support utilities for this in the Nextcloud core and we just needed to hook to these. I just ended up using a way too much time trying to figure out, why the "pre-fligth" calls from Ample were working in some cases but not in others. In the end, I figured out it was my own middleware class messing things up by attempting to do authentication checking on these pre-flight calls (which don't actually contain any user or session details). 🤦 

If you wish, you can probably monkey-patch your installation by applying the changes made in the commit linked above. It will still take some time before I'm ready to make the next official release of Music.

djtulan:
[...]

PR #1199, opened 2025-02-08
Also allow ImageController to be called from javascript

rstefko:
This fixes one more CORS error that was not covered in #1181.

### DIFF
--- a/lib/Controller/AmpacheImageController.php
+++ b/lib/Controller/AmpacheImageController.php
@@ -63,6 +63,7 @@ class AmpacheImageController extends Controller {
 	 * @PublicPage
 	 * @NoCSRFRequired
 	 * @NoSameSiteCookieRequired
+	 * @CORS
 	 */
 	public function image(?string $token, ?string $object_id, string $object_type='album') : Response {
 		if ($token === null) {


rstefko:
Could you also please add [Ample.fm](https://ample.fm) to the [Ampache clients list](https://github.com/owncloud/music/wiki/Ampache)?

paulijar:
Okay, great. But what puzzles me is that what is the case where this actually helps? Because when I'm testing Ample.fm, I can see the images being loaded without problems regardless of this CORS annotation.

And sure, I'll add it to the client list.

rstefko:
Currently it is used in the fullscreen view to get background color based on the associated image.

paulijar:
Okay. Interestingly, it seems that whenever the playing song changes, Ample queries the image.php twice, once with argument `thumb=32` and second time with `thumb=10`. The implementation on ownCloud Music ignores this argument but I guess it's intended to adjust the size of the returned image. Now, first of these calls succeeds and the seconds one fails with the CORS problem. The difference of these calls seems to be that the failing request contains the Origin header while the other request doesn't.
kuva

rstefko:
Thank you.

In most cases images are loaded from HTML elements that are rendered using Svelte by modifying HTML DOM, which does not require CORS. But this specific case is loading image directly in javascript code.

paulijar:
This is now released in Music v2.1.2.

PR #1222, opened 2025-05-11
Fix a typo

noantiq:
Hi, while reading through the settings I found a typo that I simply searched for and replaced here.


paulijar:
Thanks!

PR #1241, opened 2025-08-10
Implement play queue sync for Subsonic

### LINKED ISSUE #1103: Support synced play queue

mattwellss:
- Store user play queue in `oc_preferences`
- support for music tracks and podcast episodes
- resolves #1103 
- builds on #1104 (credit to original author jplitza)

Original pull request includes discussion of limiting the response size of `getPlayQueue`, which isn't implemented at this point, but it's worth discussing


lib/Controller/SubsonicController.php
```diff
@@ -1067,17 +1071,49 @@ protected function deleteBookmark(string $id) : array {
...
+		foreach ($playQueue['entry'] as &$entry) {
+			[$type, $id] = self::parseEntityId($entry);
+			if ($type === 'track') {
+				$entry = $this->trackToApi(
+					$this->trackBusinessLayer->find($id, $this->user())
+				);
+			} else if ($type === 'podcast_episode') {
+				$entry = $this->podcastEpisodeBusinessLayer->find($id, $this->user())->toSubsonicApi();
+			}
+		}
```
mattwellss, line 1090: I wonder about this section. If an `$entry` somehow doesn't match either `"track"` or `"podcast_episode"`, it'll be left in the response as an invalid element. Are there any other types of media that might end up saved in `savePlayQueue`? If this is a reasonable concern (and I dunno if it is), I can solve it by sanitizing the input on `savePlayQueue` or by generating a new array of just the valid items from `$playQueue['entry']` in `getPlayQueue`

paulijar: That's a good question. The third kind of media playable on the Subsonic clients is a radio station. I'm not certain if any client puts those to the play queue, though. The API would allow also video playback, but this is not supported in the Music app (at least for now or in the near future).

But yeah, it would probably be safer to construct the list so that only the supported types of entities would be retained.

paulijar:
Sorry for the delay, I have been busy with other things but now I managed to do some testing with this.

Overall, I think this works pretty much as it should. There's one minor deviation from the Subsonic XML schema https://www.subsonic.org/pages/inc/api/schema/subsonic-rest-api-1.16.1.xsd as there the attribute `username` is defined as required but it's missing from the current implementation here. I would guess that no client actually uses this attribute for anything but better to follow the schema, still.

However, there's also one major issue now and that's the performance. The execution time of `getPlayQueue` now grows linearly with the number of tracks on the queue. For example, with 100 tracks in the queue, I saw execution time to be 25 seconds or more which is way too much. The reason for this is most certainly that there is call to `$this->trackToApi` for each track of the queue. This will cause several SQL queries to be executed per each track and all that adds up quickly.

The good news is that this is a kind of problem which has been already solved elsewhere in the Subsonic API. There exists the function `SubsonicController::tracksToApi` which does the conversion for multiple tracks at once, using clever tricks to minimize the number of SQL queries. So the function `getPlayQueue` should now be reorganized so that all the `Track` objects would be fetched first from `TrackBusinessLayer` and then all of them would be converted at once to the Subsonic API format. What makes this non-trivial is the fact that there may be podcast episodes interlaced among the songs. So likely, we would need to construct the result in multiple steps. It could be something like this:
```
$entries = \array_map(fn($entry) => self::parseEntityId($entry), $playQueue['entry']);
$trackEntries = \array_filter($entries, fn($entry) => $entry[0] === 'track');
$tracks = \array_map(fn($entry) => $this->trackBusinessLayer->find($entry[1], $this->user(), $trackEntries);
$tracks = ArrayUtil::createIdLookupTable($tracks); // track IDs will become keys
$apiTracks = $this->tracksToApi($tracks); // this preserves the keys of the array

$entries = \array_map(function($entry) {
	[$type, $id] = $entry;
	if ($type === 'track') {
		return $apiTracks[$id];
	} else if ($type === 'podcast_episode') {
		return $this->podcastEpisodeBusinessLayer->find($id, $this->user())->toSubsonicApi();
	}
}, $entries);
```
Caveat: I didn't try to run this, syntax errors are possible.

While at it, one thing to consider is what will happen in case the queue contains an ID which doesn't match any item in the library. This could happen for example if the item has been deleted since the queue was saved. At the moment, the action `getPlayQueue` will fail with the Subsonic error "entity not found". This will happen when `BusinessLayer::find` throws a `BusinessLayerException` and it's not caught within `SubsonicController`. It could be nicer if only the invalid entity was omitted but other parts of the queue would be returned. Of course, the case shouldn't be terribly common so the problem is quite minor.

mattwellss:
Thanks for the review! I appreciate the testing/feedback and I'll work on those improvements. Unsure if I'll get more commits pushed until next weekend, though

lib/Controller/SubsonicController.php
```diff
@@ -1067,17 +1071,64 @@ protected function deleteBookmark(string $id) : array {
...
 	 */
-	protected function getPlayQueue() : array {
-		// TODO: not supported yet
-		return ['playQueue' => []];
+	protected function getPlayQueue() : Response {
+		/** @var array|false $playQueue */
+		$playQueue = json_decode($this->configManager->getUserValue($this->user(), $this->appName, 'play_queue', 'false'), true);
+
+		if (!$playQueue) {
+			return $this->subsonicResponse([]);
```
mattwellss, line 1079: I'm returning an empty response rather than an empty playqueue when we haven't saved one. Seems better than generating metadata for a canned response... but I couldn't find anything in the API docs regarding how to handle this situation, so maybe it's noncompliant

lib/Controller/SubsonicController.php
```diff
@@ -1067,17 +1071,64 @@ protected function deleteBookmark(string $id) : array {
...
+
+		$playQueue['entry'] = \array_reduce(
+			$parsedEntries,
+			function ($pqEntries, $parsedEntry) use ($apiTracks) {
+				[$type, $id] = $parsedEntry;
+				if ($type === 'track' && isset($apiTracks[$id])) {
+					$pqEntries[] = $apiTracks[$id];
+				} else if ($type === 'podcast_episode') {
+					try {
+						$pqEntries[] = $this->podcastEpisodeBusinessLayer->find($id, $this->user())->toSubsonicApi();
```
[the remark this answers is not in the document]
paulijar: Yes, I think that would be good

lib/Controller/SubsonicController.php
```diff
@@ -1067,17 +1071,64 @@ protected function deleteBookmark(string $id) : array {
...
+			'entry' => array_filter(
+				$id,
+				fn (string $entityId) => in_array(self::parseEntityId($entityId)[0], ['track', 'podcast_episode'])
+			),
+			'changedBy' => $c,
+			'position' => $position,
+			'current' => $current,
+			/** @see Util::formatZuluDateTime (if only we could pass a datetime!) */
+			'changed' => $changedDateTime->format('Y-m-d\TH:i:s.v\Z'),
+			'username' => $u
```
paulijar, line 1125: We should use `$this->user()` here instead of the URL argument `$u`. The reason is that the Music app supports the OpenSubsonic extension [API Key Authentication](https://opensubsonic.netlify.app/docs/extensions/apikeyauth/) which means that the client may pass the argument `apiKey` instead of `u` and `p`. The function will return the correct username in either case.

Another small difference is that the URL argument `$u` is case insensitive but `$this->user()` always contains the "proper" casing i.e. the same format in which the username was given when the account was created.

paulijar:
Performance-wise, the latest version works great with local songs: While the 100-song queue previously took 25+ seconds, now I can get a 500-song queue in less than a second. 👍

lib/Controller/SubsonicController.php
```diff
@@ -1067,17 +1071,79 @@ protected function deleteBookmark(string $id) : array {
...
+					\array_map(
+						fn ($entry) => $entry[1],
+						\array_filter($parsedEntries, fn ($parsedEntry) => $parsedEntry[0] === $type)
+					),
+					$this->user()
+				))
+			)],
+			array_keys($typeLookupMap),
+			array_values($typeLookupMap)
+		));
```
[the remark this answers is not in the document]
paulijar: I appreciate the functional programming style but it is true that this is a bit difficult to read or at least requires a lot of focusing to understand it correctly. As a rule of thumb, having more than two level of nesting in control structures is a bit of a code smell, and the same could probably be applied on these `array_*` functions which are essentially just higher-level control structures. Sometimes the deep nesting might be the best option but at least it's a sing that considering also other design options could be fruitful.

So if you can make it any easier to read with intermediate vars, then that's good. But if not, then that's not a blocker for merging. I'll give this another test round on Sunday or next week. If no problems are found, I'll merge this, unless you want to still change something.

I'm planning to release Music v2.3.0 rather soon, and I'd like to get this merged before that. The deadline for the release is the end of September, before NC32, but preferably I'd like to do it during the next two weeks.

paulijar:
This is now released in Music v2.3.0.

PR #1245, opened 2025-08-22
translation: update the Hungarian translation

rkrisztian:


rkrisztian:
Please note that I found it difficult to translate expressions like "Search for [tracks]" accurately, because in Hungarian the most appropriate translation would be "Keresés számokra," but "számokra" (for tracks)  is a conjugated form of "számok" (tracks), so if "tracks" is used in a context without the same conjugation, then the translation will be inaccurate in that other context. To fix this, for now I used the expression "Keresés erre: [számok]", which basically means "Search for this: [tracks]", hoping that the other contexts "tracks" is used for needs no conjugation, but then the PO Editor complains that I ended the translation of "Search for" with a colon. Let me know if you want me to correct this in any way.

paulijar:
Thanks for the PR! I kind of anticipated that the design of the Advanced search view might be tricky for some translations but your work-around sounds good enough for now. I'll take a closer look on this on Sunday or next week.

We don't need the binary `music.mo` file for anything so you can safely delete it.

rkrisztian:
Hi, @paulijar, OK, I deleted it. Oh I almost forgot: should I do something with the `hu_HU` folder? Ideally it would be the same as `hu`, because Hungarian is the de facto official language of Hungary, and there are no significantly different dialects or regional variations officially recognized, the language code alone (`hu`) is often sufficient for us. But at least the translations shouldn't differ between those locale codes.

rkrisztian:
> As you have now added the translations to the `hu` folder, the direction of that copying should be reversed. You can do it or I can do it later.

No problem, I will do it.

paulijar:
I now tested the translation in practice and no obvious issues were found. One minor thing is that while I understand your rationale of using the colon to translate "Search for", I think that the same probably doesn't apply to the two other similar complaints on Poedit:
image

The "{{artist}} appears on albums" is used in the artist details pane when the artist is featured on some album:
image

The "Similar to" is used in the artist details pane when the Last.fm connection is available. Here, the Hungarian translation actually doesn't quite fit in the provided space but the full title can be seen when hovering the mouse cursor above the title.
image

rkrisztian:
> One minor thing is that while I understand your rationale of using the colon to translate "Search for", I think that the same probably doesn't apply to the two other similar complaints on Poedit:

Thanks, I should have done more thorough testing during my translations. My bad. Corrected.

paulijar:
It's actually not used for similar *albums* but for similar *artists*. If this doesn't change the significant parts of your reasoning, then "Hasonló(k)" is fine for me.

paulijar:
Now thinking about it one more time, it's even a bit unnecessary to use the parentheses here for the plural suffix. In practice, if Last.fm has any information about some artist, then it always returns multiple "similar artists". The definition of "similar" here may sometimes be quite loose. On the Music UI, this doesn't depend on the contents of the user library since we list here both in-library artists and artists not found in the library. So presumably, the singular case is something which never happens.

paulijar:
Merged. Thanks for all your work!

PR #1266, opened 2025-11-21
Implement Last.fm scrobbling

### LINKED ISSUE #339: [feature request] scrobble songs to last.fm
it would be nice to have songs scrobbled to last.fm in the same way Spotify does.

mattwellss:
Resolves #339 
- When configured, sends _scrobbles_ to Last.fm
- Adds new `config.php` keys _music.scrobble_api_key_, _music.scrobble_api_secret_, _music.scrobble_api_service_
- Updates settings UI to add buttons for creating a [session](https://www.last.fm/api/webauth)
- Adds a new route for handling a token response callback from Last.fm
- Adds another new route for clearing an existing Last.fm session
- Updates the Track Business Layer to become an emitter (emits scrobble events)

Screenshots!

## Settings update overview:
Settings view; ready to create a user Last.fm session

## Authentication flow:
https://github.com/user-attachments/assets/e33e3d8f-2562-4002-b1b7-c9dd7e86f7ed

## Not configured view:
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 ## [Unreleased]
 
 ### Added
+- Scrobbling to Last.fm (requires opt-in from the admin and the user)
+  [#1266](https://github.com/owncloud/music/pull/1266) @mattwellss
 - Context menu with "Save playlist" function for advanced search results
   [#1256](https://github.com/owncloud/music/issues/1256)
 - Searching for radio stations with the advanced search


webpack.config.js
```diff
@@ -22,7 +22,8 @@ module.exports = {
   entry: {
     app: './js/index.app.js',
     dashboard_music_widget: './js/index.dashboard.js',
-    files_music_player: './js/index.embedded.js'
+    files_music_player: './js/index.embedded.js',
+    scrobble_getsession_result: './js/index.getsessionresult.js'
```
paulijar: There's no file index.getsessionresult.js included in the PR. Is this unnecessary or has the file been forgotten?

mattwellss:
Some stuff I was considering that didn't make it in (but might be considered blockers)

- Storing scrobbles in a new DB table to be sent on a cron, decoupling the scrobble service's availability/performance from the `scrobble` call's performance
- Implementing more scrobble backends. I spent a bit of time on ListenBrainz (https://listenbrainz.readthedocs.io/en/latest/users/api-compat.html) but kept hitting `410 Gone` responses. I *think* that's because of some broad IP blocking they do but I gave up on it pretty quick
- Building an admin-level UI for configuring API key/secret. Would allow encrypting the secret, which is currently impossible afaik

lib/Controller/ScrobblerController.php
```diff
@@ -0,0 +1,89 @@
...
+ * ownCloud - Music app
+ *
+ * This file is licensed under the Affero General Public License version 3 or
+ * later. See the COPYING file.
+ *
+ * @author Morris Jobke <hey@morrisjobke.de>
+ * @author Pauli Järvinen <pauli.jarvinen@gmail.com>
+ * @copyright Morris Jobke 2013, 2014
+ * @copyright Pauli Järvinen 2019 - 2025
+ */
```
paulijar: You should add you own name as author and copyright owner when creating new files. I will add mine later if I ever make modifications to the file. When modifying existing files, you may add your name there after the existing names but that's up to you.

I have used such such a logic with years of the copyright that the first year is the creation date and the end year is the time of the latest change.

js/app/controllers/views/settingsviewcontroller.js
```diff
@@ -322,6 +322,33 @@ angular.module('Music').controller('SettingsViewController', [
 			});
 		};
 
+		$scope.generateSession = function() {
```
paulijar: This function and $scope.clearSession should be renamed to contain the term "scrobble" or "scrobbling" to make it obvious to any reader, what they are about

lib/Controller/ScrobblerController.php
```diff
@@ -0,0 +1,87 @@
...
+		} catch (\Throwable $t) {
+			$success = false;
+			$headline = 'Failed to authenticate.';
+			$getSessionResponse = $t->getMessage();
+			$instructions = 'Authentication failure. Please review the error message and try again.';
+		} finally {
+			return new StandaloneTemplateResponse($this->appName, 'scrobble-getsession-result', [
+				'lang' => $this->l10n->getLanguageCode(),
+				'success' => $success,
+				'headline' => $this->l10n->t($headline),
```
paulijar: The l10n->t(...) function cannot be used like this because the scraper script is not smart enough to find the translatable strings through variable names. So whenever the function is used, the translatable string should be inlined in the function call: `$l10n->t('Some string to translate')`.

lib/Service/ScrobblerService.php
```diff
@@ -0,0 +1,234 @@
...
+			);
+			$this->config->setUserValue($userId, $this->appName, 'scrobbleSessionKey', $encryptedKey);
+		} catch (\Throwable $e) {
+			$this->logger->error('Unable to save session key ' . $e->getMessage());
+			throw $e;
+		}
+	}
+
+	public function getApiKey() : ?string {
+		return $this->config->getSystemValue('music.scrobble_api_key', null);
```
paulijar: We already have the settings key `music.lastfm_api_key` so having another for the same purpose doesn't seem necessary. And if we later add support for Libre.fm, then we could add another key `music.librefm_api_key`. Then, ideally, the admin could configure even both keys at the same time, and each user would be allowed to select the scrobbling service they prefer.

To lay ground work for supporting multiple services like described above, also `music.scrobble_api_secret` should be `music.lastfm_api_secret`. Also `music.scrobble_api_service` is unnecessary since the availability of a scobbling service can be deduced from the availability of the API key *and* API secret.

lib/Service/ScrobblerService.php
```diff
@@ -0,0 +1,234 @@
...
+			return false;
+		}
+
+		$timestamp = $timeOfPlay->getTimestamp();
+		$scrobbleData = \array_merge($this->generateBaseMethodParams('track.scrobble'), [
+			'sk' => $sessionKey,
+		]);
+
+		/** @var array<Track> $tracks */
+		$tracks = $this->trackBusinessLayer->findById($trackIds);
```
paulijar: Also `userId` should be passed to `findById`. Otherwise we may leak one user's data to other users of the same cloud.

templates/partials/views/settingsview.php
```diff
@@ -178,6 +178,16 @@ class="icon-delete reset-button" id="reset-podcasts" ng-click="resetPodcasts()"/
 			</dl>
 		</div>
 	</div>
+	<h2 translate>Scrobbler Integration</h2>
+	<div>
+		<p ng-hide="settings.scrobbler.configured" translate>The server admin can configure your server for scrobbling. The <strong>Admin</strong> section below details how to achieve this.</p>
+		<p ng-show="settings.scrobbler.configured" translate >The server admin has configured your server for scrobbling.</p>
+		<p ng-show="settings.scrobbler.configured && !settings.scrobbler.hasSession">Click the <strong>Get Scrobble Session</strong> button to send your streams to {{settings.scrobbler.service}}.</p>
+		<p ng-show="settings.scrobbler.configured && settings.scrobbler.hasSession">Click the <strong>Clear Session</strong> to stop scrobbling.</p>
```
paulijar: When there is a session, the text should state clearly something like "Your plays are being scrobbled to Last.fm".

paulijar (changes requested):
Overall, it's a very good PR, and with some improvements on the commented topics, it will be excellent.

lib/AppInfo/Application.php
```diff
@@ -189,4 +197,29 @@ private static function hlsEnabled(IConfig $config, ?string $userId) : bool {
...
 		return $enabled;
 	}
+
+	/**
+	 * @param mixed $context On Nextcloud, this is \OCP\AppFramework\Bootstrap\IRegistrationContext.
+	 *                       On ownCloud, this is \OCP\AppFramework\IAppContainer.
+	 */
+	private function registerScrobblerServices($context) : void
+	{
+		$context->registerService('scrobblerServices', function () {
```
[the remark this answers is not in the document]
paulijar: I guess the problem is now that scrobblerServices depend on the TrackBusinessLayer and TrackBusinessLayer depends on the scrobblerServices so we have a circular dependency.

The architecture of the Music app is such that `Db` is the lowest layer, `BusinessLayer` is built on top of that, then `Service`, and `Controller` is the highest layer. Normally, the higher layers depend on the lower layers but not vice versa. Now `TrackBusinessLayer` calling the `ScrobblerService` works against this pattern.

To maintain the usual layered structure, it would probably be best to reorganize the scrobbling so that the Controller modules now calling `TrackBusinessLayer::recordTrackPlayed` would instead call a corresponding method on the Service-level scrobbler class. That one should then call `TrackBusinessLayer::recordTrackPlayed`. That way, the BusinessLayer wouldn't need to know anything about the Service layer.

I'd also say that mostly the Controller modules shouldn't be aware that there may be several scrobbling services. That is, there should be only single scrobbling interface class which gets injected to `SubsonicController`, `AmpacheController`, and `MusicApiController`. This interface module can then dispatch the scrobbling command to various provider-specific classes or instances (Last.fm, Libre.fm, ...).

lib/Controller/AmpacheController.php
```diff
@@ -14,6 +14,7 @@
 
 namespace OCA\Music\Controller;
 
+use OCA\Music\Service\Scrobbler;
```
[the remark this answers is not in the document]
mattwellss: I'll add it to the group with the other `OCA\Music\Service` classes

mattwellss:
> Most of the issues reported by Scrutinizer could still be addressed. The one about CurlHandle we can ignore for now. But for example for the return values of curl_exec, we know for fact that the way it is used, it cannot return true. We could hint that to Scrutinizer for example by writing

Appreciate the push. I realized there's no reason to expose the use of `curl` outside the function which uses it, meaning the type change is no longer relevant. That change plus your suggestion of using `assert` cleaned up the curl-related issues.

At this point, the curl calling functionality of `ExternalScrobbler` can be extracted into a separate class. Doing so would unlock unit testing, which wasn't possible before.

Scrutinizer is now complaining about `$xml ?: null` (ensure a nullable `SimpleXMLElement` return type for `execRequest`). PHP's docs say `simplexml_load_string` can return false, but Scrutinizer doesn't recognize that. :shrug: 

Final inspection note: Scrutinizer doesn't like my use of `catch (\TypeError` when user id or token is missing (https://scrutinizer-ci.com/g/owncloud/music/inspections/ab5b74ff-f2c7-4a85-b70d-c5295c572578/issues/files/lib/Controller/ScrobblerController.php?status=new&orderField=path&order=asc&honorSelectedPaths=0). This approach definitely works (https://3v4l.org/uSJIX#v7.4.32), but I guess it's a little outside the norm.

paulijar:
> I realized there's no reason to expose the use of curl outside the function which uses it, meaning the type change is no longer relevant. That change plus your suggestion of using assert cleaned up the curl-related issues.

Definitely much better now. But now, if `curl_exec` would return `false`, the execution would be terminated instead of returning a `null` or throwing like in other error cases. So maybe we should instead write it like this:
```
	$xmlString = \curl_exec($ch);
	$xml = \is_string($xmlString) ? \simplexml_load_string($xmlString) : null;
	return $xml ?: null;
```

> Scrutinizer is now complaining about `$xml ?: null` (ensure a nullable SimpleXMLElement return type for execRequest). PHP's docs say `simplexml_load_string` can return false, but Scrutinizer doesn't recognize that.

Maybe it would help Scrutinizer if an explicit phpdoc type-hint was added `/** @var SimpleXMLElement|null $xml */`. But if not, then we will just live with it.

> Final inspection note: Scrutinizer doesn't like my use of catch (\TypeError when user id or token is missing (https://scrutinizer-ci.com/g/owncloud/music/inspections/ab5b74ff-f2c7-4a85-b70d-c5295c572578/issues/files/lib/Controller/ScrobblerController.php?status=new&orderField=path&order=asc&honorSelectedPaths=0). This approach definitely works (https://3v4l.org/uSJIX#v7.4.32), but I guess it's a little outside the norm.

I have been struggling myself with how the type checking for URL arguments should be made in the Controller interface and I still haven't find a solution I would always be happy to apply. For one thing, often passing the arguments is mandatory for the meaningful operation of the REST endpoint. But if we mark the parameters as non-nullable, then trying to call the endpoint without proper arguments is responded with "500 Internal server error" which isn't very nice. It probably also gets logged on the server as an error. Something like "400 Bad request" would be must more correct, but to achieve that, we need to mark the parameters as nullable and then null-check manually and return the error response. That's a bit tedious if done extensively on all API endpoints, but maybe that would still be the way to go.

In case of `ScrobblerController`, also the argument `$serviceIdentifier` could be technically omitted and probably should be nullable in the function signature.

Another similar problem is the dependency injected `$userId`. It should be non-null always when there is a logged-in user. But sometimes the Controller may get instantiated in a situation where the user session has already been terminated. In that case, there would be an ugly error logged if the $userId was not nullable. I *think* that the actual endpoint methods should not get called in situations like this.

lib/Controller/ScrobblerController.php
```diff
@@ -0,0 +1,126 @@
...
+								array $externalScrobblers) {
+		parent::__construct($appName, $request);
+		$this->l10n = $l10n;
+		$this->userId = $userId;
+		$this->appName = $appName;
+		$this->externalScrobblers = $externalScrobblers;
+	}
+
+	/**
+	 * @PublicPage
```
paulijar: This and `clearSession` shouldn't have the `@PublicPage` annotation because that indicates an endpoint which can be called without a logged-in user (like on a public share page). Instead, there should be `@NoAdminRequired` to allow calling for any logged-in user (the default is to allow only admin users). And in case you wander, the entry points of `AmpacheController` and `SubsonicController` are marked as "public" because both of these have their own user credential handling, independent from the Nextcloud user sessions.

The `@NoSameSiteCookieRequired` is probably necessary here as written, since this endpoint is called by the Last.fm site. But in the function `clearSession`, it shouldn't be needed.

mattwellss:
> In my opinion, it would be cleaner to have a null-check block for this than rely on the `TypeError` and it would also make Scrutinizer happy. And it wouldn't really need almost any more code.

Gotcha, I'll update the controller to proactively handle invalid `$token` values.

> Also the possible null `$this->userId` is still nagged on by Scrutinizer. This could be null-checked with a separate block, or just silenced with the null coalescing: `$scrobbler->generateSession($token, $this->userId ?? '');`.
> 
> In the latter case, if the code ever got executed with the null $userId (which is doubtful), the result would probably be `InvalidArgumentException` which is already caught.

This might not change your recommendation, but I checked this and forgot to report back. When not logged in, all controller actions in NextCloud get routed to the login page with a `redirect_url` param. No clue what OwnCloud does, though, so it's probably better to take steps to avoid potentially logging non-useful errors.

paulijar:
@mattwellss As you probably noticed, I rebased and merged the PR yesterday. Thanks again for your excellent work! It was a long haul. But that's nothing compared to the decade which the feature request sat on the back-log waiting for someone to pick it up ;).

Btw, I did the small scrobbling-related updates to the wiki pages already on Sunday.

I'm somewhat busy until Christmas but I hope I can get the Music v2.5.0 out still this year, including also this feature.

